PR #4, opened 2019-04-15
Implementation of parsing and evaluation of Witness nodes.

roconnor-blockstream:
* Also adds a `skip` field to nodes in the type DAGs that points to sort of non-trivial grandchild.  See the `typeSkip` function for details.  This optimization mitgates a DOS attack where the attacker builds a type by repeatedly taking a product with ONE and a non-trivial type in an attempt to increase the time of witness data processing without increasing the size of the witness data.


C/dag.c
```diff
@@ -136,6 +136,61 @@ static sha256_midstate wmrIV(int32_t tag) {
...
+ * finalize the SHA-256 computation by adding SHA-256 padding and set 's' to the resulting SHA-256 hash.
+ *
+ * Precondition: uint32_t s[8];
+ *               uint32_t block[16];
+ *               NULL != count;
+ */
+static void sha256_end(uint32_t* s, uint32_t* block, const size_t len) {
+  size_t count = len;
+  sha256_pushBit(s, block, &count, 1);
+  if (448 < count % 512) {
```
apoelstra: To be clear - the use of `<` here rather than `<=` is simply wrong, because in the `==` case the bit you just pushed will be overwritten by the length?

(I notice you fixed this in a later commit.)

C/dag.c
```diff
@@ -268,3 +325,94 @@ void forceJets(dag_node* dag, const analyses* analysis, const size_t len, JET_FL
     }
   }
 }
+
+/* This function fills in the 'WITNESS' nodes of a 'dag' with the data from 'witness'.
+ * For each 'WITNESS' : A |- B expression in 'dag', the bits from the 'witness' bitstring are decoded in turn
+ * to construct a compact representation of a witness value of type B.
+ * This funciton only returns 'true' when exactly 'witness.len' bits are consumed by all the 'dag's witness values.
```
apoelstra: Typo `funciton`

C/dag.c
```diff
@@ -268,3 +325,94 @@ void forceJets(dag_node* dag, const analyses* analysis, const size_t len, JET_FL
...
+ * Precondition: dag_node dag[len] and 'dag' without witness data and is well-typed with 'type_dag';
+ *               witness is a valid bitstring;
+ *
+ * Postcondition: dag_node dag[len] and 'dag' has witness data and is well-typed with 'type_dag'
+ *                  when the result is 'true';
+ */
+bool fillWitnessData(dag_node* dag, type* type_dag, const size_t len, bitstring witness) {
+  for (size_t i = 0; i < len; ++i) {
+    if (WITNESS == dag[i].tag) {
+      if (witness.len <= 0) {
```
apoelstra: I think `<=` vs `==` is confusing, since `witness.len` is a `size_t` and cannot be below 0

apoelstra:
ACK. But fair warning, this is a *lot* of really intricate code and it made my head hurt to read.

C/dag.c
roconnor-blockstream: Do you think so?  I prefer `<=` to offer protection against refactorings that might change the signedness of these sort of types.  Why might the signedness change?  Because I'm told that C will somehow take advantages of UB in signed arithmetic to optimize loops, and this is something I haven't really considered yet.  Perhaps in the future I will want to change the `size_t` in loops like on line 343 to `int` and cause a ripple effect of causing preferences of `int` over `size_t` in `witness.len`.

However, if you feel strongly, please let me know, and I will change it to `==`

C/dag.c
roconnor-blockstream: I believe the use of `<` is correct here.  If the count is exactly 448 then message length is 447 bits, and only a [message that is 448 bits or larger is padded beyond the block size](https://crypto.stackexchange.com/a/29851/47684) requires running another iteration of the compression function.

The bit is not overwritten because after writing the 448th bit, there are still 64 bits in the 512 bit block, which is enough space to write the length suffix without overwriting the bit.

roconnor-blockstream:
Thanks. It is now intricate because parsing of the witness data can only proceed after type inference is completed.  This means I now have a DAG data structure that goes through several phases of having partial data: untyped, typed, witnessed, and if you mix them up, everything will go wrong.  However, given the limited type system of C, it's a bit hard to enforce that everything progresses in the proper order.

The good news is that I think this is as bad as it gets.  Adding primitives shouldn't be too hard.  Supporting disconnect shouldn't make things too much worse.  And ultimately the API for this library should boil down to a single function that parses, executes, and returns the boolean result (and probably the weight determined by the static analysis as well).

C/dag.c
roconnor-blockstream: Fixed.

C/dag.c
roconnor-blockstream: However, if this were an error, I do want to fix it, even if it is only part of history.  I have a specific commit that is intended as only an optimization, not a change in functionality.

C/dag.c
apoelstra: No, this seems reasonable. `<=` is fine.

C/dag.c
apoelstra: Ah! Yep, I missed that `count` is one more than the message length.

apoelstra:
ACK

PR #7, opened 2019-06-04
Fix typos and remove unnecessary trailing whitespace

ysangkok:
I didn't standardize US/UK spelling because I don't know which one you prefer.


roconnor-blockstream:
Thanks for the corrections!  As a Canadian, I've been trying to standardize on Canadian spellings; however, I never know what that is. :)  Maybe I need to bookmark http://www.lukemastin.com/testing/spelling/cgi-bin/database.cgi?database=spelling

PR #10, opened 2019-06-24
Dune/Proof-General usage instructions

ysangkok:
If somebody wants to understand the Coq code, currently they'd have to learn how to take the Nix result and launch Emacs/Proof-General inside that built environment. I think it would be valuable to provide build instructions for users who are not using Nix, because it is quite a hurdle to learn Nix just to understand this code. By providing build instructions and package files for building with Dune (which is easier to install and use than Nix), more people will be able to understand Simplicity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,16 @@ Software artifacts can be built using [Nix](https://nixos.org/nix/).
 To build the Haskell project, run `nix-build -A haskell`.
 To build the Coq project, run `nix-build -A coq`.
 
+To build the Coq part without Nix, you can try:
+
+1. download and extract VST from the URL in `vst.nix`
+1. in the VST directory, do: `make compcert/lib/Integers.vo sha/SHA256.vo sha/functional_prog.vo`
+1. in the root Simplicity directory, do:
+1. `dune build Coq/Util`
+1. `dune build Coq/Simplicity`
+1. Proof General won't find the built `vo` files unless they are moved from the directory where Dune puts them: `for i in $(find _build/default/Coq -name '*.vo' ); do mv -v $i $(echo $i | cut -d/ -f3-); done`
+1. Now you can use Proof General to browse the proofs.
+
 ## Documentation
 
 Detailed documentation can be found in the `Simplicity-TR.tm` TeXmacs file.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -17,6 +17,16 @@ Software artifacts can be built using [Nix](https://nixos.org/nix/).
 To build the Haskell project, run `nix-build -A haskell`.
 To build the Coq project, run `nix-build -A coq`.
 
+To build the Coq part without Nix, you can try:
+
+1. download and extract VST from the URL in `vst.nix`
+1. in the VST directory, do: `make compcert/lib/Integers.vo sha/SHA256.vo sha/functional_prog.vo`
```
roconnor-blockstream: I'd like to instead add a step here that says something like "export the VST directory into the COQPATH environment variable" and then remove all references to VST-2.3 from the `_CoqProject` and `dune` files.

_CoqProject
```diff
@@ -0,0 +1,10 @@
+-R VST-2.3/compcert/lib compcert.lib
+-R VST-2.3/compcert/flocq compcert.flocq
+-R VST-2.3/compcert/x86_32 compcert.x86_32
+-R VST-2.3/compcert/common compcert.common
+-R VST-2.3/sha sha
+-R VST-2.3/msl VST.msl
+-R VST-2.3/veric VST.veric
+-R VST-2.3/floyd VST.floyd
+-R Coq/Util Util
```
roconnor-blockstream: I think we should prefer to use `-Q` instead of `-R` for these flags, which is what I use in `Simplicity.Coq.nix`.

roconnor-blockstream:
Is is okay to move `_CoqProject` into the `Coq` directory?

roconnor-blockstream:
I cannot but help wonder if you wouldn't find it easier to use `_CoqProject` file with `coq_makefile` versus `dune` at this point in time?

ysangkok:
Let's first decide if Dune should be used at all. My problem with using just `_CoqProject` and `coq_makefile` is that I just cannot seem to find a working a file. I tried listing every single `.v` file in _CoqProject but that doesn't seem to cause `coq_makefile` to generate targets for them. Since `coq_makefile` is haphazardly designed, I prepared this PR using Dune. But of course, if there is a `_CoqProject` file that works with `coq_makefile` and with emacs, that would be much simpler.

roconnor-blockstream:
I've created https://github.com/ElementsProject/simplicity/pull/13 as an alternative to this PR.  @ysangkok are you willing to try out the instructions in that branch?

PR #11, opened 2019-07-23
Implementation of Elements primitives.

roconnor-blockstream:
See Appendix A of the [Simplicity Tech Report](https://github.com/ElementsProject/simplicity/blob/pdf/Simplicity-TR.pdf) for details about the specification of the Elements primitives.

This includes a simple "checkSigHashAll" testcase with a test transaction data.

The project now builds a `elementsSimplicity.a` shared library to be linked with Elements


C/typeInference.c
```diff
@@ -649,9 +649,10 @@ type* mallocTypeInference(dag_node* dag, const size_t len, const combinator_coun
     , { .isBound = true, .bound = { .kind = PRODUCT, .arg = { &bound_var[5], &bound_var[5] } } }
     , { .isBound = true, .bound = { .kind = PRODUCT, .arg = { &bound_var[6], &bound_var[6] } } }
     , { .isBound = true, .bound = { .kind = PRODUCT, .arg = { &bound_var[7], &bound_var[7] } } }
+    , { .isBound = true, .bound = { .kind = PRODUCT, .arg = { &bound_var[8], &bound_var[8] } } }
     };
-  /* 'bound_var[8]' is bound to the type TWO^256. Eight non-trivial bindings were made. */
-  size_t bindings_used = 8;
+  /* 'bound_var[9]' is bound to the type TWO^256. Nine non-trivial bindings were made. */
```
apoelstra: Is this supposed to say `TWO^512`, or is this change incorrect? I believe it is true that `bound_var[i]` is bound to type `TWO^i` for all `i`.

roconnor-blockstream: For all `0 < i` we have that `bound_var[i]` is bound to the type `TWO^(2^(i-1))`.  FWIW, this all gets rewritten more clearly in
https://github.com/ElementsProject/simplicity/pull/11/commits/5430fae78b91561d9e66592ecf6858dde5b069e6

apoelstra: After offline discussion I realized I was confusing `TWO^i` with `TWO^(2^i)`, which is a very different number

C/deserialize.h
```diff
@@ -69,23 +38,30 @@ int32_t decodeUptoMaxInt(bit_stream* stream);
...
- *                  (*witness) represents a some bitstring;
- *                if 'result < 0' or 'wintess->len == 0' then witness->arr == NULL;
+ * Postcondition: if the function returns '0'
+ *                  '*witness' represents a some bitstring
+ *                   and '*allocation' points to allocated memory sufficient for at least the 'witness->arr' array.
+ *                       (Note that '*allocation' may be 'NULL' if 'n == 0')
+ *                if the function returns a negative value then '*allocation == NULL'
  */
-int32_t decodeMallocWitnessData(bitstring* witness, bit_stream* stream);
+int32_t decodeMallocWitnessData(void** allocation, bitstring* witness, bitstream* stream);
```
apoelstra: I think it'd be a cleaner API if there were a function that took a `bitstring*` and freed its memory, rather than returning the allocation through an outptr and requiring the user free it themselves.

roconnor-blockstream: I did try that design but there is a major problem in that some `bitstring`s own their memory and (most) other `bitstring`s share their memory.  The solution means that `bitstrings` all share their memory and someone else owns it.

There is also a minor problem that `bistring`s have pointers to `const` and thus cannot be freed.  While not a major problem by itself, it does guide the solution in this direction.

C/unreachable.h
```diff
@@ -0,0 +1,17 @@
+/* This module defines an UNREACHABLE macro that calls '__builtin_unreachable' if it exists.
+ * Placing UNREACHABLE after a branch can allow an optimizer to eliminate the branch.
+ */
+#ifndef UNREACHABLE
```
apoelstra: I wonder if we should require a `#define` to make this actually map to `__builtin_unreachable`, and have it map to a "output to stderr then abort" otherwise. `__builtin_unreachable` is scary because it's actually UB to call, so if it's ever used mistakenly then crazy things will happen.

roconnor-blockstream: FWIW, some inspection of the output of GCC 8 (in a variant of this code) shows that using a `__builtin_unreachable` does successfully optimise code by removing evaluation on conditions that lead to unreachable code.

It's probably fair to perform a `DEBUG` check here to write a message to stderr, but otherwise as it stands in an `NDEBUG` build will already do crazy undefined things if invariants get violated.

roconnor-blockstream: Filed https://github.com/ElementsProject/simplicity/issues/12 to rework this and the assertions to keep safety during production builds.

C/typeInference.c
```diff
@@ -310,7 +311,7 @@ static bool typeInference(unification_arrow* arrow, const dag_node* dag, const s
         }          } };
       APPLY_BINDING(&(arrow[dag[i].child[0]].source), &((binding)
         { .kind = PRODUCT
-        , .arg = { &(arrow[i].source), word256Type }
+        , .arg = { &(arrow[i].source), &(bound_var[9]) }
```
apoelstra: Can't we replace this `9` with `word256` (I guess, we'd need to promote `enum TypeNamesForJets` from primitives.c to primitives.h.)

It then seems to me that the dozen+ comments referring to 9 by value can go away.

C/primitive/elements/primitive.c
```diff
@@ -8,6 +8,113 @@
...
+  one     = 0,
+  two     = 1,
+  word2   = 2,
+  word4   = 3,
+  word8   = 4,
+  word16  = 5,
+  word32  = 6,
+  word64  = 7,
+  word128 = 8,
+  word256 = 9, /* The typename for word256 must be at index 9 */
```
roconnor-blockstream: Have mallocBoundVars return the offset where word256 is located rather than fixing it to the number 9.

C/primitive/elements.c
```diff
@@ -0,0 +1,424 @@
...
+#include <string.h>
+#include "elements/primitive.h"
+#include "../deserialize.h"
+#include "../eval.h"
+#include "../sha256.h"
+#include "../typeInference.h"
+
+#define PADDING(alignType, allocated) ((alignof(alignType) - (allocated) % alignof(alignType)) % alignof(alignType))
+
+/* Add a 256-bit hash to be consumed in big endian order by an ongoing SHA-256 evaluation.
```
apoelstra: Prefer we drop the phrase "big endian order", since hashes don't have endianness, only conversions from hashes to numbers have endianness. You can say "without reversing it" if you want to emphasize that there is no satoshi crap going on with this function.

C/include/simplicity/elements.h
```diff
@@ -88,4 +88,24 @@ typedef struct transaction transaction;
  * Precondition: NULL != rawTx
  */
 extern transaction* mallocTransaction(const rawTransaction* rawTx);
+
+/* Deserialize a Simplicity program from 'file' and execute it in the environment of the 'ix'th input of 'tx'.
+ * If the file isn't a proper encoding of a Simplicity program, '*success' is set to false.
+ * If EOF isn't encountered at the end of decoding, '*success' is set to false.
+ * If 'cmr != NULL' and the commitment Merkle root of the decoded expression doesn't match '*wmr' then '*success' is set to false.
```
apoelstra: Typo `*wmr` should be `*cmr`

I also think writing `*cmr` to mean "the 8 words pointed at by `cmr`" is a bit confusing, but I'm not sure what a better notation would be.

apoelstra:
ACK except for above comments.

C/include/simplicity/elements.h
roconnor-blockstream: Done.

C/primitive/elements.c
roconnor-blockstream: Done.

C/primitive/elements/primitive.c
roconnor-blockstream: Done.

C/typeInference.c
roconnor-blockstream: Done.

C/Makefile
```diff
@@ -13,15 +15,26 @@ jetTable.c: jetTable.gperf
...
+	$(CC) -c $(CFLAGS) $(CWARN) -Wno-unused-parameter $(CPPFLAGS) -o $@ $<
+
+primitive/elements/jets.o: primitive/elements/jets.c
+	$(CC) -c $(CFLAGS) $(CWARN) -Wno-unused-parameter -Wno-switch-enum -Wswitch $(CPPFLAGS) -o $@ $<
+
 %.o: %.c
 	$(CC) -c $(CFLAGS) $(CWARN) $(CPPFLAGS) -o $@ $<
 
-test: test.o dag.o deserialize.o eval.o frame.o hashBlock.o jets.o jetTable.o schnorr1.o schnorr8.o type.o typeInference.o
+elementsSimplicity.a: bitstream.o dag.o deserialize.o eval.o frame.o jets.o jetTable.o sha256.o type.o typeInference.o primitive/elements.o primitive/elements/jets.o primitive/elements/primitive.o
```
roconnor-blockstream: @apoelstra Should this be named `libelementssimplicity.a` instead?

apoelstra:
Fixes look good!

PR #14, opened 2019-09-11
Represent null confidential commitments as a NULL pointer.

roconnor-blockstream:
Given how Elements' CConfidentialCommitment internal data structure works, it is easier to use a NULL pointer for NULL confidential commitments.

We also now explicitly allow issuance.blindingNonce and assetEntropy to be NULL pointers for a NULL issuance.
This allows us set issuance = {0} for NULL issuances.
This was alreayd implicitly allowed, so this bit is only a documentation change.


roconnor-blockstream:
Will bundle other changes in a later PR.

PR #15, opened 2019-09-13
Minor ammendments to the Elements API

roconnor-blockstream:
* Null commitments are now represented as `NULL` pointers.
* Uniformly use `unsigned char` over `uint8_t`.
* Merkle roots are passed via an `unsigned char[32]` array instead of the `uint32_t[8]` array.


apoelstra:
utACK

apoelstra:
re-utACK

PR #18, opened 2019-10-29
secp256k1 jets

roconnor-blockstream:
Review C/secp256k1/README.md for information about our copy of libsecp256k1.

Also included is a commit that eliminates unused parameter warnings.


apoelstra:
What commit did you get these libsecp files from? and I guess you renamed `secp256k1.c` to `secp256k1_impl.h`, and moved the schnorrsig files out from the `modules/` tree to the root?

apoelstra:
oh, I see, you actually modified all the files

C/jets-secp256k1.c
```diff
@@ -0,0 +1,218 @@
...
+}
+
+static struct {
+   secp256k1_ecmult_context ctx;
+   char alloc[SECP256K1_ECMULT_CONTEXT_PREALLOCATED_SIZE];
+   bool initialized;
+} ecmult_static;
+
+/* This function will initialize the 'ecmult_static' global variable if it hasn't already been initialized. */
+static void ensure_ecmult_static(void) {
```
apoelstra: The thread-unsafety of this really makes me uncomfortable, though I can't think of any other way to do this. Maybe putting a secp context into `txEnv`?

C/secp256k1/eckey.h
```diff
@@ -0,0 +1,24 @@
...
+#include <stddef.h>
+
+#include "group.h"
+#include "scalar.h"
+#include "ecmult.h"
+
+static int secp256k1_eckey_pubkey_parse(secp256k1_ge *elem, const unsigned char *pub, size_t size);
+static int secp256k1_eckey_pubkey_serialize(secp256k1_ge *elem, unsigned char *pub, size_t *size);
+
+#if 0
```
apoelstra: Is this `#if 0` code meant to be committed?

C/secp256k1/ecmult_impl.h
```diff
@@ -0,0 +1,680 @@
...
+/** Fill a table 'pre' with precomputed odd multiples of a.
+ *
+ *  There are two versions of this function:
+ *  - secp256k1_ecmult_odd_multiples_table_globalz_windowa which brings its
+ *    resulting point set to a single constant Z denominator, stores the X and Y
+ *    coordinates as ge_storage points in pre, and stores the global Z in rz.
+ *    It only operates on tables sized for WINDOW_A wnaf multiples.
+ *  - secp256k1_ecmult_odd_multiples_table_storage_var, which converts its
+ *    resulting point set to actually affine points, and stores those in pre.
+ *    It operates on tables of any size, but uses heap-allocated temporaries.
```
apoelstra: FWIW this comment was updated in https://github.com/bitcoin-core/secp256k1/pull/677 -- apparently at your advice.

C/secp256k1/secp256k1_impl.h
```diff
@@ -0,0 +1,139 @@
...
+        if (ret) {
+            *outputlen = len;
+        }
+    }
+    return ret;
+}
+
+/* Converts the point encoded by a secp256k1_pubkey into its absolute value,
+ * i.e. keeps it as is if it is positive and otherwise negates it. Sign is set
+ * to 1 if the input point was negative and set to 0 otherwise. */
```
apoelstra: This comment (and the function implementation) are also different upstream. See https://github.com/jonasnick/secp256k1/blob/schnorrsig/src/secp256k1.c#L736

apoelstra: And in particular, I find the existing text which talks about a point being "positive" or "negative" to be confusing

C/jets-secp256k1.c
roconnor-blockstream: This is a more general problem with the entire library.  I've created [a separate issue](https://github.com/ElementsProject/simplicity/issues/33) to track it .

roconnor-blockstream:
Moved to https://github.com/ElementsProject/simplicity/pull/34.

PR #23, opened 2019-11-13
Correct type inference for OutputNullDatum primitive.

roconnor-blockstream:
The correct output type for OutputNullDatum is S (S ((TWO^2 * TWO^256) + (TWO + TWO^4))).
This error was probably introduce by implementing type inference for an older design of the OutputNullDatum primitive and not updating it when the design was redone.

### DIFF
--- a/C/primitive/elements/primitive.c
+++ b/C/primitive/elements/primitive.c
@@ -33,10 +33,11 @@ enum TypeNamesForJets {
   sWord256,
   sSWord256,
   sWord32,
+  word2TimesWord256,
   twoPlusWord4,
-  pubKeyPlusBitPlusWord4,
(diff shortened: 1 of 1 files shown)


roconnor-blockstream:
This fixes all the assertion failures found by gmaxwell @ https://lists.ozlabs.org/pipermail/simplicity/2019/000019.html

gmaxwell:
Not that I'm qualified to review this, but -- I confirm the jet implementation appears to be writing the type the comment says it is and the same type now appears to be described by the unification_vars.  Also I confirm none of my tests assert anymore and no new crashes found with a bit of additional fuzzing.

apoelstra:
ACK

PR #26, opened 2019-11-15
Fix various heap management errors.

roconnor-blockstream:
1. All *malloc* functions have post-conditions that claim that the returned pointer is set to NULL if the function fails.
However, 'decodeMallocWitnessData' and 'mallocTypeInference' were not setting pointers to NULL under various failure conditions.

2. In functions that do local heap allocations, we cannot use early returns because we bypass the call to 'free' and leak memory.
'elements_simplicity_execSimplicity' had some early returns that occured after the first heap allocation ('decodeMallocDag').

My manual search of the codebase didn't uncover any further occurences of the above kind of errors.

### DIFF
--- a/C/deserialize.c
+++ b/C/deserialize.c
@@ -382,6 +382,7 @@ int32_t decodeMallocDag(dag_node** dag, combinator_counters* census, bitstream* 
  *                if the function returns a negative value then '*allocation == NULL'
  */
 int32_t decodeMallocWitnessData(void** allocation, bitstring* witness, bitstream* stream) {
+  *allocation = NULL;
   int32_t witnessLen = getBit(stream);
   if (witnessLen < 0) return witnessLen;
   if (0 < witnessLen) witnessLen = decodeUptoMaxInt(stream);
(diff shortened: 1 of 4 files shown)


gmaxwell:
This looks correct for me and passes my tests.

gmaxwell:
Causes "execSimplicity of elementsCheckSigHashAllTx1 on tx1 failed" in test.c.  Also test.c:141 if !dag above, perhaps initialize void* witnessAlloc at the top of the function.

roconnor-blockstream:
Oops, I didn't include `test.c` in my review.

roconnor-blockstream:
Added fixes to `test.c`.

gmaxwell:
During the execution of stock test.c in elements test,

primitive/elements.c       if (*success && EOF != getc(file)) {  

*success can be uninitialized: len<0 true, err < 0 false,

roconnor-blockstream:
Fixed.

gmaxwell:
This now tests good and looks reasonable to me.

gmaxwell:
Still tests clean and looks good to me.

apoelstra:
ACK. tested clean in valgrind.

PR #28, opened 2019-11-16
Fix issues with zero sized VLAs

roconnor-blockstream:
To address issue https://github.com/ElementsProject/simplicity/issues/25 we begin our removal of VLAs starting with all VLAs used in `sizeof`.

We also fix an issue with possible zero sized allocations for 'extra_var' within `mallocTypeInference`.


gmaxwell:
This looks right and tests out okay for me.

apoelstra:
ACK from me too. A bit hard to follow the logic changes but I think the final result is followable (as much as allocation logic can be) on its own.

PR #30, opened 2019-11-16
Avoid calling exit for unimplemented functionality.

roconnor-blockstream:
gmaxwell asked if we could avoid calling `exit` to aid in program analysis.
While there still is one call to `exit` in `eval.c` it ought to be unreachable.
There is also a call to `exit` in `test.c` if `fmemopen` fails, but I don't even think that is possible.


gmaxwell:
ACK, I've been running essentially this patch (though with ERR_HIDDEN being returned) for a couple days.

apoelstra:
Works for me. ACK.

PR #40, opened 2020-01-29
Deserialize discounted jets.

roconnor-blockstream:
Current jets include 32-bit arithmetic and SHA-256 compression.


C/primitive/elements/primitive.c
```diff
@@ -188,54 +209,126 @@ static int32_t either(jetName* result, jetName a, jetName b, bitstream* stream)
  *               NULL != stream
```
roconnor-blockstream: Update the comment above replacing `prefix of '10'` with `prefix of '1'` since we are now parsing that second bit.

apoelstra:
ACK. Ran all tests with valgrind; all passed, no errors.

I notice in this PR the distinction between application-specific primitives and jets is basically removed. Does this match how you think about things (and how I should think about things?)

roconnor-blockstream:
Logically jets and primitives are still distinct, but for the C implementation that distinction is immaterial.  The main difference is in how their Merkle roots are defined, but for the C implementation, the Merkle roots are just taken as given.

PR #41, opened 2020-02-09
nix: add full simplicity dev environment via shell.nix

jb55:
This adds a full Simplicity development environment using nix-shell,
and adds some docs explaining how to use it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,9 @@ This project contains
 Software artifacts can be built using [Nix](https://nixos.org/nix/).
 
 * To build the Haskell project, run `nix-build -A haskell`.
-* To use the Haskell project, try `nix-shell -p "(import ./default.nix {}).haskellPackages.ghcWithPackages (pkgs: [pkgs.Simplicity])"`.
 * To build the Coq project, run `nix-build -A coq`.
+* For a Simplicity-Haskell development environment, type `nix-shell --arg coq false`.
+* Typing `nix-shell` will provide a full C, Coq and Haskell development environment.
 
 ### Building without Nix
 
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,33 @@
+{ haskell ? true
+, coq     ? true
+, c       ? true
+, nixpkgs ? import <nixpkgs> {}
+}:
+let
+  simplicity      = import ./. {};
+  optional        = nixpkgs.lib.optional;
+  mkDerivation    = nixpkgs.stdenv.mkDerivation;
+
+  haskellDrv =
+    let
+      haskellDevTools = pkgs: with pkgs; [cabal-install hlint hasktags];
+      haskellPkgs     = pkgs: simplicity.haskell.buildInputs ++ haskellDevTools pkgs;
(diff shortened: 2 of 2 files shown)


shell.nix
```diff
@@ -0,0 +1,33 @@
...
+  optional        = nixpkgs.lib.optional;
+  mkDerivation    = nixpkgs.stdenv.mkDerivation;
+
+  haskellDrv =
+    let
+      haskellDevTools = pkgs: with pkgs; [cabal-install hlint hasktags];
+      haskellPkgs     = pkgs: simplicity.haskell.buildInputs ++ haskellDevTools pkgs;
+      haskellDevEnv   = simplicity.haskellPackages.ghcWithPackages haskellPkgs;
+    in
+      mkDerivation {
```
jb55: for some reason I can't use `haskellDevEnv` directly here, I need to wrap it in another mkDerivation.  `ghcWithPackages` and `nix-shell` interaction continues to confuse me

roconnor-blockstream:
Are you willing to rebase this on https://github.com/ElementsProject/simplicity/tree/haskell-non-backpack?  That branch has additional options to build Bitcoin and Elements only variants of the Haskell Simplicity library.

jb55:
@roconnor-blockstream `cabal build` looks broken on that branch, I get:

```
TODO: add support for multiple packages in a directory
CallStack (from HasCallStack):
  error, called at ./Distribution/Client/ProjectOrchestration.hs:542:9 in main:Distribution.Client.ProjectOrchestration
```

it would be nice to support a standard cabal workflow instead of relying on nix builds (which aren't incremental). It looks like the only way to do that is to have the cabal files in their own respective directories. Right now I'm achieving incremental builds using haskell-interactive mode and specifying :set -iHaskell/Core paths, etc. but that's not ideal.

README.md
```diff
@@ -18,8 +18,9 @@ This project contains
 Software artifacts can be built using [Nix](https://nixos.org/nix/).
 
 * To build the Haskell project, run `nix-build -A haskell`.
-* To use the Haskell project, try `nix-shell -p "(import ./default.nix {}).haskellPackages.ghcWithPackages (pkgs: [pkgs.Simplicity])"`.
```
roconnor-blockstream: This line above enters a shell with Simplicity already built (useful for manual testing the code in ghci), and thus shouldn't be eliminated as it is unrelated to your shell which enters an environment to build Simplicity.

roconnor-blockstream:
I believe I've address the issue of having multiple `.cabal` files in one directory though a pretty big hack.  Please take note of the `shellHook`s I've added.

I think it would be best to move this on top of the `haskell-non-backpack` branch.  In that case the `haskell` parameter shouldn't be a boolean, but rather a choice between `Simplicity` `Simplicity-Bitcoin` and `Simplicity-Elements`.

Also I suggest looking into whether using `pkgs.mkShell` would simplify your code.

roconnor-blockstream:
Since the `haskell-non-backpack` branch is effectively stale, I've gone ahead and rebased and merged this PR in https://github.com/ElementsProject/simplicity/commit/7c716b951bca72e0665e1f6629322d8503d4ee51.

PR #46, opened 2020-04-07
Swap the arguments to the first parameter of disconnect.

roconnor-blockstream:
This new order allows for a `disconnect iden t` idiom which I think will be a common use case for `disconnect`.  The forthcoming "universal sighash" work uses this idiom.


roconnor-blockstream:
@apoelstra I only expect you to review the `C/typeInference.c` file.

PR #49, opened 2020-04-18
Swaps

ProofOfKeags:
This PR adds 4 simplicity expressions. They are all described by their types, I am unsure if any further testing/verification is required here.

### DIFF
--- a/Haskell/Core/Simplicity/Term/Core.hs
+++ b/Haskell/Core/Simplicity/Term/Core.hs
@@ -17,6 +17,7 @@ module Simplicity.Term.Core
  -- The string of @i@'s and @o@'s is meant to resemble a binary number that denotes an index to the leaves of a perfect binary tree.
  , oh, ih, ooh, oih, ioh, iih, oooh, ooih, oioh, oiih, iooh, ioih, iioh, iiih
  , swapP {- product swap -}, swapS {- sum swap -}
+ , cnf2dnf {- distribute -}, dnf2cnf {- factor -}
  -- * Language extensions
  , Assert(..), fail0
  , Witness(..)
--- a/Haskell/Core/Simplicity/Term/Core.hs
+++ b/Haskell/Core/Simplicity/Term/Core.hs
@@ -114,7 +114,7 @@ swapP = pair (drop iden) (take iden)
 
 -- | Term for swapping positions in sums (Commutativity of Addition): A + B |- B + A
 swapS :: (Core term, TyC a, TyC b) => term (Either a b) (Either b a)
-swapS = iden &&& unit >>> (match (take (injr iden)) (take (injl iden)))
+swapS = copair (injr iden)) (injl iden)
 
 -- | Term for distributing products over sums (product of sums "CNF" to sum of products "DNF")
 cnf2dnf :: (Core term, TyC a, TyC b, TyC c) => term (a, Either b c) (Either (a, b) (a, c))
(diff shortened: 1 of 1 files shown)


Haskell/Core/Simplicity/Term/Core.hs
```diff
@@ -105,6 +107,22 @@ iioh = drop ioh
...
+swapS :: (Core term, TyC a, TyC b) => term (Either a b) (Either b a)
+swapS = iden &&& unit >>> (match (take (injr iden)) (take (injl iden)))
+
+-- | Term for distributing products over sums (product of sums "CNF" to sum of products "DNF")
+cnf2dnf :: (Core term, TyC a, TyC b, TyC c) => term (a, Either b c) (Either (a, b) (a, c))
+cnf2dnf = swapP >>> (match (injl swapP) (injr swapP))
+
+-- | Term for factoring common products out of sums (sum of products "DNF" to product of sums "CNF")
+dnf2cnf :: (Core term, TyC a, TyC b, TyC c) => term (Either (a, b) (a, c)) (a, Either b c)
+dnf2cnf = iden &&& unit >>> match (ooh &&& injl oih) (ooh &&& injr oih)
```
roconnor-blockstream, line 124: `iden &&& unit >>> match (take (oh &&& injl ih)) (take (oh &&& injr ih))` is slightly better.

ProofOfKeags: for conciseness, performance, canonical representation?

roconnor-blockstream: Is it both more concise, which is probably the most important reason, and there is a fractional performance improvement since you only execute take once instead of twice.

roconnor-blockstream:
If you are willing to indulge me, can you define and use the following:

    copair s t = iden &&& unit >>> case (take s) (take t)

Haskell/Core/Simplicity/Term/Core.hs
```diff
@@ -16,6 +16,8 @@ module Simplicity.Term.Core
  --
  -- The string of @i@'s and @o@'s is meant to resemble a binary number that denotes an index to the leaves of a perfect binary tree.
  , oh, ih, ooh, oih, ioh, iih, oooh, ooih, oioh, oiih, iooh, ioih, iioh, iiih
+ , swapP {- product swap -}, swapS {- sum swap -}
```
roconnor-blockstream, line 19: You can remove the comments.

ProofOfKeags:
I actually was just defining this. Copair is a much better name than the
“either” I was working with. I’ll put it in.

roconnor-blockstream:
So I don't really like the names `dnf2cnf` and `cnf2dnf`.  I've been looking around to try to find better names.  Leaning towards `leftDistribute` and `leftFactor`.  But, if it is the case that we don't have any particular applications in mind for these functions, how would you feel about leaving them out for now?

ProofOfKeags:
I wouldn't mind leftDistribute and leftFactor. Before committing this, in an earlier version I called them distribute/factor. The left qualification seems good here, although they are just a `comp swapP` or a `comp swapS` away from the right variants respectively.

I'm not generally a fan of leaving them out altogether at the moment. I'm not completely opposed, but it seems like building up the combinator library, even when we don't know what they would be for is a good idea. Obviously, this can be taken too far, but these properties seem sufficiently general to me that despite no known concrete applications, it seems entirely plausible that they will come in handy in the future.

I'll defer to your preference here, they'd be easy enough to add back in if we decide we need them later. I just think the general philosophy of building up a wealth of general combinators probably helps, especially since the applications that may justify them are probably several layers of abstraction up.

ProofOfKeags:
That said, a cursory search on hoogle reveals no functions with the following types

(a, Either b c) -> Either (a, b) (a, c)
Either (a, b) (a, c) -> (a, Either b c)

So they may perhaps be useless after all

roconnor-blockstream:
Indeed, I suspect it is rarely used in practice.

PR #50, opened 2020-05-05
Implementing disconnect

roconnor-blockstream:
Examples that exercise 'disconnect' will follow this PR (unless the reviewer demands otherwise).

### DIFF
--- a/C/dag.c
+++ b/C/dag.c
@@ -129,50 +129,47 @@ static sha256_midstate wmrIV(tag_t tag) {
   UNREACHABLE;
 }
 
-/* Given a well-formed dag representing a Simplicity expression, compute the commitment Merkle roots of all subexpressions.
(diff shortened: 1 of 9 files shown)


apoelstra:
I feel a bit weird about using `SIZE_MAX` as a limit in consensus code. Woudn't it be better to use `2^31-1` or something? Realistically we are never going to have 2 billion of anything in a program that can be executed in a blockchain context.

roconnor-blockstream:
Indeed that is why I have

    /* :TODO: replace this check with a consensus critical limit. */

Perhaps you would prefer that I file a github issue to make the issue more prominent?

apoelstra:
Yes please :)

roconnor-blockstream:
This has now been noted as part of issue https://github.com/ElementsProject/simplicity/issues/54.

apoelstra:
Also verified a bunch of calculations in section 6.1 of the TR, as per Russell's request

PR #60, opened 2020-08-17
Fix potential out-of-bounds shifting errors in sha256.c

roconnor-blockstream:
Various amendments to ensure that when shifting the type being shifted upon is guaranteed to have an adequate number of bits.

Fixes https://github.com/ElementsProject/simplicity/issues/59


roconnor-blockstream:
@real_or_random are you able to review?

PR #61, opened 2020-08-28
Allow sha256_init to set output to arbitrary array.

roconnor-blockstream:
Rather than only a sha256 midstate.
This lets us hash directly into one half a 64-byte block which is useful for many Merkle tree constructions.


roconnor-blockstream:
This is in preparation for switching to BIP-340+ style tagged hashes.

apoelstra (approved):

PR #62, opened 2020-08-28
New merkle roots

roconnor-blockstream:
- Changes Merkle Root definitions to use BIP-340+ style tagged hashes.
- Replaces Witness Merkle Root with Identiy Merkle Roots for identifying "identical" nodes in a dag (not implemented in C yet).
- Renames Witness Merkle Root to Annotated Merkle Root (now only for debugging purposes).


Simplicity-TR.tm
```diff
@@ -2298,6 +2298,34 @@
...
+
+  <\eqnarray*>
+    <tformat|<table|<row|<cell|SHA256<rsup|t><rsub|<2>><around*|(|x|)>>|<cell|\<assign\>>|<cell|SHA256<rsub|<2>><around*|(|\<mu\><rsup|\<ast\>><around*|(|<around*|(|\<iota\><rsup|<2><rsup|8>><rsub|<2><rsup|\<ast\>>>|)><rsup|\<ast\>><around*|(|p\<cdummy\>p|)>|)>\<cdummy\>x|)>>>|<row|<cell|SHA256<rsup|t><rsub|<2><rsup|8>><around*|(|x|)>>|<cell|\<assign\>>|<cell|SHA256<rsub|<2><rsup|8>><around*|(|p\<cdummy\>p\<cdummy\>x|)>>>|<row|<cell|>|<cell|where>|<cell|p\<assign\>BE<rsub|256><around*|(|SHA256<rsub|<2><rsup|8>><around*|(|t|)>|)>>>>>
+  </eqnarray*>
+
+  \ We can compute the SHA-256 midstate after compressing this first block
+  containing the prefix tag.
+
+  <\equation*>
+    SHA256<rsup|t><rsub|IV>\<assign\>SHA256<rsub|Block><around*|\<langle\>|SHA256<rsub|IV>,\<Delta\><around*|(|SHA256<rsub|<2><rsup|8>><around*|(|t|)>|)>|\<rangle\>>
```
sanket1729, line 2356: Should there be another SHA256 before the \Delta pair?

sanket1729:
Also verified the amr for elements schnorr in https://github.com/ElementsProject/simplicity/blob/0624dd723850969c084492a4f86bd86bbcfb8112/C/primitive/elements/checkSigHashAllTx1.c

Simplicity-TR.tm
roconnor-blockstream: `SHA256[Block]` has type `2^256 * 2^512 -> 2^256`.  `SHA256[2^8](t)` has type `2^256`.  The Delta function duplicates its argument so  `\Delta(SHA256[2^8](t))` has type `2^256 * 2^256` which is equal to `2^512`.  This is the type required at that position.

sanket1729: Ah, I thought `\Delta` infix operator for pair. So, in case of a single argument, It duplicates it arg.

roconnor-blockstream: Right.  The `\Delta` function creates a pair of duplicates while the infix `white triangle` operator pairs up the results of two functions.  I can see how the notation might be confusing, although it is (somewhat) standard.

`\Delta := [[iden \whitetriangle iden]]`

roconnor-blockstream:
I've renamed the prefix from `Simplicity` to `Simplicity-Draft` so all the IMRs for the jets need to be updated, and the AMRs and CMRs double checked again.

apoelstra (approved):
utACK.

Did not look at the Coq code; skimmed the Haskell; read the C in some detail and ran the tests but I didn't replicate Sanket's computation of the new hashes.

roconnor-blockstream:
Thanks.  I think I'll wait for @sanket1729 to rerun his CMR / AMR computations with the amended prefix.

sanket1729:
@roconnor-blockstream @apoelstra  verified the new CMR/AMR computations for hashblock, schnorr0 and schnorr6 and elements sighash

PR #65, opened 2020-10-08
Avoid macros beginning with 'E'.

roconnor-blockstream:


real-or-random:
Thank god the project name does not start with an "E".

apoelstra (approved):

PR #66, opened 2020-11-10
Change the order of the inputs to full adders and full subtractors.

roconnor-blockstream:
There is a minor improvement of well-aligned frames when the carry bit parameter comes first.
Because frames are aligned with their end, an extra bit at the beginning of the frame will overflow into its own word, when we are adding two machine word sized values.
Prior to this change the carry bit is at the end, and would push every word over by one bit when adding two machine word sized values.
(Note that "optional" 'S Word' parameters also have the bit that specifies if the optional parameter exists at the beginning.  It was for this reason that frames were aligned at the end, so that the bit specifying whether the optional parameter exists ends up in the low order bit of its own word, like a boolean flag.)


roconnor-blockstream:
@apoelstra please review the C code.

apoelstra (approved):
ack 2b2bc3cf9327aefbb866184040856e17b6d54926

Reviewed the C code and ran the unit tests in valgrind. Did not verify the actual Simplicity code in the tests.

PR #67, opened 2021-01-18
Implement radix sort for 256-bit hash values.

roconnor-blockstream:
The 'hasDuplicates' function is not yet used by the Simplicity interpreter, however it will be used to ensure that every node's identity Merkle root is unique.

### DIFF
--- a/C/rsort.h
+++ b/C/rsort.h
@@ -1,0 +1,43 @@
+#ifndef SIMPLICITY_RSORT_H
+#define SIMPLICITY_RSORT_H
+
+#include <limits.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+#include "sha256.h"
+
+_Static_assert(UCHAR_MAX < SIZE_MAX, "UCHAR_MAX >= SIZE_MAX");
+#define CHAR_COUNT ((size_t)1 << CHAR_BIT)
+
+/* Internal function used by 'hasDuplicates'.  Do not call directly. */
+const sha256_midstate* rsort(size_t* scratch, const sha256_midstate** a, size_t len, size_t level);
+
+/* Searches for duplicates in an array of pointers to 'sha256_midstate's.
+ * If 'true' is returned, '*duplicates' is set according to whether duplicate 'sha256_midstate' values were found.
+ * If 'true' is returned, the array is also wiped, with all pointers set to NULL.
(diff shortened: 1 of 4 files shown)


C/rsort.h
```diff
@@ -0,0 +1,43 @@
...
+
+#include "sha256.h"
+
+_Static_assert(UCHAR_MAX < SIZE_MAX, "UCHAR_MAX >= SIZE_MAX");
+#define CHAR_COUNT ((size_t)1 << CHAR_BIT)
+
+/* Internal function used by 'hasDuplicates'.  Do not call directly. */
+const sha256_midstate* rsort(size_t* scratch, const sha256_midstate** a, size_t len, size_t level);
+
+/* Searches for duplicats in an array of pointers to 'sha256_midstate's.
```
apoelstra, line 17: nit spelling

C/rsort.h
```diff
@@ -0,0 +1,43 @@
...
+static inline bool hasDuplicates(bool* duplicates, const sha256_midstate** a, size_t len) {
+  _Static_assert(sizeof((*a)->s) * CHAR_BIT == 256, "sha256_midstate.s has unnamed padding");
+  _Static_assert(sizeof((*a)->s) < SIZE_MAX / CHAR_COUNT, "CHAR_BIT is way too large");
+  _Static_assert((sizeof((*a)->s) + 1) * CHAR_COUNT <= SIZE_MAX/sizeof(size_t), "sizeof(size_t) is way too large");
+  size_t * scratch = malloc((sizeof((*a)->s) + 1) * CHAR_COUNT * sizeof(size_t));
+
+  if (scratch) {
+    *duplicates = rsort(scratch, a, len, sizeof((*a)->s));
+
+    /* Censor the array because its current order is implemenation defined. */
```
apoelstra, line 35: Is this necessary even in the case that `duplicates` is false? In this case `rsort` should complete, and since there are no duplicates there is no question of what the correct ordering should be.

apoelstra: Ah ok, you're refusing to commit to the order. Fair enough.

apoelstra (approved):
ack dca6c012858d30bc3940d366bc6476ab7f5bef70

C/rsort.h
roconnor-blockstream: For the record, even in the case of the `rsort` completing the order is implementation defined because it sorts differently on little endian versus big endian architectures.  This is not something we want to expose in consensus code.

PR #69, opened 2021-04-09
Verify cannonical DAG order.

roconnor-blockstream:
Manually tested by exchanging the `/*Check first child. */` and `/*Check second child. */` fragments and seeing that the check fails.


C/dag.c
```diff
@@ -371,6 +371,105 @@ void computeAnnotatedMerkleRoot(analyses* analysis, const dag_node* dag, const t
...
 
+/* Verifies that the 'dag' is in canonical order, meaning that nodes under the left branches have lower indices than nodes under
+ * right branches, with the exception that nodes under right braches may (cross-)reference identical nodes that already occur under
+ * left branches.
+ *
+ * Returns 'true' if the 'dag' is in canonical order, and returns 'false' if it is not.
+ *
+ * May modify dag[i].aux values.
+ *
+ * Precondition: dag_node dag[len] and 'dag' is well-formed.
```
apoelstra, line 382: I think you should also have a precondition that neither `sourceType` or `targetType` have been initialized at this point, since this method will erase them through the anonymous usion with `aux`.

roconnor-blockstream: That isn't really a logical preconditions since it isn't needed to entail the post condition.  And while we don't have to strive for the weakest precondition in general, the statement that such and such field is uninitialized I don't think is actually something that is part of a formal language of conditions about the state of variables in C, as there is no way to distinguish uninitialized fields from initialized fields.

Ultimately these conditions are intended to stand in place of later VerifiableC formal pre and post condtions, and those conditions, of course, will capture what you are getting at, that any source of target type from the input isn't preserved, which is captures by, more or less, the absence of the statement that it is preserved.

That all said, I'm happy to expand upon the dag[i].aux comment to point out that it also means that any sourceType and targetType values may be erased.

apoelstra (approved):
ack 6088e1805904a7be66b7edc2491d60609c59ef19 aside from nit about preconditions

As always, it took me a very long time to grok the algorithm but in the end it is very clever and efficient.

C/dag.c
apoelstra: Agreed with all points. I'd appreciate updating the (informal description part of the) comment to warn users to call this prior to type-checking.

apoelstra:
Ran all unit tests with valgrind; tried the "swap child[0] and child[1]" trick and confirmed that it caused a bunch of test failures.

C/dag.h
```diff
@@ -99,7 +99,12 @@ static inline size_t numChildren(tag_t tag) {
 typedef struct dag_node {
   jet_ptr jet;
   sha256_midstate cmr;
-  size_t sourceType, targetType;
+  union {
+    size_t aux; /* Used as scratch space for verifyNoDuplicateIdentityRoots. */
```
roconnor-blockstream, line 103: Should reference `verifyCanonicalOrder`.

PR #70, opened 2021-06-21
Refactoring in preparation for secp256k1 jets.

roconnor-blockstream:
Beyond rearranging and renaming a few things, this PR merges in a new jet encoding format that makes room for a wide range of new jets, including EC jets.  It also updates the signature implementation to the modern and final BIP-0340 version.


apoelstra (approved):
ACK 56dcabefc155c1960d98d479bf339e6593d86920

PR #71, opened 2021-06-29
C: clean up build a bit

apoelstra:
Cleans up the `Makefile` so that things will buid without needing to change CFLAGS and LDFLAGS.

It's still not great; you still need to manually build `libsha256compression/` before building `C/` and you still need to specify `LD_LIBRARY_PATH=../libsha256compression` before running the test binary, but it's progress.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,10 @@ Simplicity-TR.pdf
 Haskell/*.pdf
 *.eps
 result-doc
+
+*.a
+*.log
+*.o
+*.so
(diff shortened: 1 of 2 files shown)


roconnor-blockstream:
I do want to eliminate libsha256compression.  It just isn't pulling its own weight.

roconnor-blockstream:
Let's review  #72 first and then come back to this.

apoelstra:
Closing in favor of #72

PR #72, opened 2021-07-02
Remove libsha256compression

roconnor-blockstream:


apoelstra (approved):
ack 751afed1174801dd5470bc95be60359795352411

This is great! I can just run `make check` now with no additional work!

PR #77, opened 2022-01-07
Elements rangeproof

roconnor-blockstream:
Add new primitives that support access to (hashes of) range proofs and surjection proofs.  Adds these hashes to the default signature hash.


C/primitive/elements/jets.c
```diff
@@ -274,6 +300,28 @@ bool input_issuance_token_amt(frameItem* dst, frameItem src, const txEnv* env) {
   return true;
 }
 
+/* input_issuance_asset_proof : TWO^32 |- S (S TWO^256) */
```
apoelstra, line 303: In d2063b4938c2ecdddd1c855ae930e46cd2a5ce22:

This comment should have `ONE` as the source rather than `TWO^32` (same with the immediately following function).

apoelstra: Actually this same mistake occurs in the auto-generated bindings and in the Haskell -- I can't remember now what data the generated code is produced from, but I guess the typo is in there.

roconnor-blockstream: The input is the index of the input to access.  As you can see below

    uint_fast32_t i = read32(&src);

apoelstra:
Done reviewing the C code in 982367c5049c6cbc92557f03fa5b3f70f135d3f9. Only issue is the source typing thing.

Was relieved to find that this doesn't actually implement any rangeproof logic, only treats them as raw bytes to be hashed.

C/primitive/elements/jets.c
```diff
@@ -482,6 +544,22 @@ bool current_issuance_token_amt(frameItem* dst, frameItem src, const txEnv* env)
   return true;
 }
 
+/* current_issuance_asset_proof : TWO^32 |- S TWO^256 */
```
roconnor-blockstream, line 547: Oh I see.  @apoelstra your comment is supposed to be here.

C/primitive/elements/jets.c
roconnor-blockstream: Fixed in the other location.

C/primitive/elements/jets.c
roconnor-blockstream: Fixed.

C/primitive/elements/primitiveJetNode.inc
```diff
@@ -461,6 +485,18 @@
 , .sourceIx = ty_u
 , .targetIx = ty_mspbw256w64
 }
+,[CURRENT_ISSUANCE_ASSET_PROOF] =
+{ .tag = JET
+, .jet = current_issuance_asset_proof
+, .sourceIx = ty_w32
```
apoelstra, line 491: These should also be changed to be ones rather than w32s. (This and the entry immediately following it.)

Haskell/Elements/Simplicity/Elements/Primitive.hs
```diff
@@ -61,6 +64,8 @@ data Prim a b where
   CurrentIssuanceEntropy :: Prim () (S Word256)
   CurrentIssuanceAssetAmt :: Prim () (S (Conf Word64))
   CurrentIssuanceTokenAmt :: Prim () (S (Conf Word64))
+  CurrentIssuanceAssetProof :: Prim Word32 (S Word256)
```
roconnor-blockstream, line 67: Source of @apoelstra  bug.

C/primitive/elements/primitiveJetNode.inc
roconnor-blockstream: Fixed.

Haskell/Elements/Simplicity/Elements/Primitive.hs
roconnor-blockstream: Fixed.

roconnor-blockstream:
I've made some changes to attempt to prevent the same sort of bug from reoccurring.  The problem is somewhat hard because the naive specification of these primitives was still satisfied by the erroneous implementation.

apoelstra (approved):
ACK 001dd2a37c67577724e8cb2e61919046b04e9cb2

Neat. I wonder if there was any potential for abuse from the freedom of the primitives' input types. Not that I can think of.

roconnor-blockstream:
@sanket1729 I still don't know how to add you as a reviewer.  I don't know what github's criterion is.

apoelstra:
I explicitly added Sanket to this repo. The existing list of potential reviewers appears to be the ElementsProject "Owners" and nobody else.

roconnor-blockstream:
@apoelstra the primitives themselves always fix their types.  It's just that the reference implementation did not, which is why it allowed me to make an error on the input type being `Word32` instead of `()`.

C/primitive/elements/jets.c
```diff
@@ -139,6 +139,32 @@ static void issuanceTokenAmt(frameItem* dst, const assetIssuance* issuance) {
   }
 }
 
+/* Write an optional confidential asset rangeproof from an 'assetIssuance' to the 'dst' frame, advancing the cursor 257 cells.
```
sanket1729, line 142: nit: (here and ahead) 
```diff
- Write an optional confidential asset rangeproof
+ Write an optional confidential asset rangeproof hash

File: C/primitive/elements/jets.c [hunk 6]
```diff
@@ -122,32 +119,52 @@ static void issuanceEntropy(frameItem* dst, const assetIssuance* issuance) {
...
+ *
+ * Precondition: '*dst' is a valid write frame for 257 more cells;
+ *               NULL != issuance;
+ */
+static void issuanceAssetProof(frameItem* dst, const assetIssuance* issuance) {
+  if (writeBit(dst, NO_ISSUANCE != issuance->type && is_confidential(issuance->assetAmt.prefix))) {
```
roconnor-blockstream, line 148: Here and below, it looks like you may be inconsistently handling the non-confidential case.  It seems sometimes you write none in an optional type, but in other places you write a hash of the empty string.

C/primitive/elements/jets.c
```diff
@@ -382,6 +421,28 @@ bool output_null_datum(frameItem* dst, frameItem src, const txEnv* env) {
   return true;
 }
 
+/* output_surjection_proof : TWO^32 |- S TWO^256 */
```
roconnor-blockstream, line 424: output_surjection_proof and output_range_proof are wrong and have the wrong types.

The specification requires that they return `S (S TWO^256)`.

So either the specification needs to be changed or the implementation needs to be changed.

roconnor-blockstream:
While I've pushed a correction, I've reverted this to draft until I get sufficient testing infrastructure in place to detect this mistake in the [elements-rangeproof-broken](https://github.com/ElementsProject/simplicity/tree/elements-rangeproof-broken) branch.

C/primitive/elements/jets.c
roconnor-blockstream: fixed.

C/primitive/elements/primitive.h
```diff
@@ -138,6 +147,8 @@ typedef enum issuanceType {
 
 /* In Elements, a trasaction input can optionally issue a new asset or reissue an existing asset.
  * This structure contains data about such an issuance.
+ * 'assetRangeProofHash' is the SHA-256 hash of the asset amount's range proof when is_confidential(assetAmt.prefix).
+ * 'tokenRangeProofHash' is the SHA-256 hash of the token amount's range proof when is_confidential(tokenAmt.prefix).
  *
  * Invariant: If 'type == NEW_ISSUANCE' then 'contractHash' and 'tokenAmt' are active;
```
roconnor-blockstream, line 153: The list of active fields needs to be updated.

C/primitive/elements/jets.c
roconnor-blockstream: The specification has been now changed.

C/primitive/elements/jets.c
roconnor-blockstream: This now consistently returns the hash of the empty string.

apoelstra (approved):
ACK 1e391dd875f4a6a2679f740d81aed62a9f240548

I spent a while reconsidering whether we want rangeproofs in the "all-outputs hash" but I think this is the correct default. Users are of course able to construct their own arbitrary hashes if they don't like this.

PR #79, opened 2022-02-26
View null amounts in issuances as explicit zeros.

roconnor-blockstream:
This brings Simplicity in line with elements taproot OP_INSPECTINPUTISSUANCE.


roconnor-blockstream:
This is the first patch from https://github.com/ElementsProject/simplicity/pull/77 which you already reviewed before.

apoelstra (approved):
ACK c43c2597edbfb2be09db34b0ec02b942cfbe7548

Only checked the C code; skimmed the rest.

PR #81, opened 2022-03-09
Fix output_null_datum primitive

roconnor-blockstream:
The previous code would hit the assert(false) statement even after going through a successfull non-default case.


apoelstra (approved):
ACK 35f8c7e8718364902491d74e216e7813eb7fe14f

PR #82, opened 2022-03-11
Test primitives

roconnor-blockstream:
Adds randomized testing in Haskell of the C primitives via new bindings to Haskell.


apoelstra (approved):
ACK 734828f5fa6ba9b17ab5979c3a1693843ca0f286 (checked C changes only)

Confirmed from the C side that this is a move-only PR.

apoelstra (approved):
ACK c85e52b80ed80abdabc50fe34df06cd5ea90bd3a

(Checked `git range-diff` to confirm that only the commit messages changed.)

PR #90, opened 2022-05-09
Jets for issuance entropy and asset IDs.

roconnor-blockstream:
Only the code in C needs to be reviewed.


apoelstra (approved):
ACK e2e2fb4783f7b124d748e3fe256f6255862395d7

Read only the C code.

Am a tiny bit nervous about reading data into sha2 midstates' buffers but not updating the length field, but it looks like the length is never used in this code so it's alright.

roconnor-blockstream:
The `sha256_midstate` type is simply a 256-bit buffer.  Other stuff is in the `sha256_context` type.

PR #92, opened 2022-05-13
Update README

chlewe:
Adds instructions on how to build the C and Haskell projects in Linux distributions such as Debian.


apoelstra (approved):
Nice! 9b2fa6e3d68a3674341746bd541da18971e60141 looks good to me. Did not try out the haskell or coq instructions.

PR #95, opened 2022-06-10
Update tests to use new 32-bit jets.

roconnor-blockstream:
This ought to have been included in https://github.com/ElementsProject/simplicity/pull/94.


apoelstra (approved):
ACK 1157624fcd9dbb155d1ba616edafae8accbd8876

PR #96, opened 2022-06-17
Rename issuance primitives.

roconnor-blockstream:
All issuances are on the input, so no need to make names longer than necessary.


apoelstra (approved):
ACK 4457e087f98a4330b333d2a2b79fbfd8f5fd2c28

PR #98, opened 2022-07-13
Sort generated jet code by jet name.

roconnor-blockstream:
Previously the order was (implicitly) sorted by the CMR.
Sorting by name will prevent the generated code from jumping around when jet definitions are tweaked.


apoelstra (approved):
ACK e353ca3479fd3cfcddc21b565f5f3ba07f1a0a43

PR #99, opened 2022-07-15
Create a copy of (most) primitives as jets

roconnor-blockstream:
Later the primitives will be removed from the serialization format.

Notes:

- `issuance_asset_amt` and `issuance_token_amt` are being renamed `issuance_asset_amount` and `issuance_token_amount` in their jet form.  Both names are being supported temporarily.
- `*_amount` don't have associated jets for the primitives.  Instead `*_asset_amount` jets have been created that return both the asset and the amount.  The `*_amount` primitives are deprecated and will be removed.
- `annex_hash` is a deprecated primitive and will be removed.
- `fee` is deprecated and will be replaced by a jet.
- `inputs_hash` and `outputs_hash` are deprecated.  They will be superseded by a new set of sighash jets.

Only the C code requires review.


apoelstra (approved):
ACK 02f46c5cdf09c4d84e923b60f6b491630762c7ce

PR #100, opened 2022-07-19
New primitives

roconnor-blockstream:
Adds primitives for
- annexes on all inputs
- script sig hashes for all inputs
- genesis hash of the chain

Also replaces the isPegin primitive with a pegin primitive that (optionally) returns the parent genesis hash of the pegged in funds.


apoelstra (approved):
ACK 9f28d91d54dd0d23a168ca053af91862ce70c5b3

PR #102, opened 2022-08-03
New sighash

roconnor-blockstream:
Adds a series of jets to build up a new sigall hash function.


roconnor-blockstream:
@apoelstra This will require significant review.  Because you ought to review the sighash structure itself.  You can either try to read the C code or read the Haskell code to understand the data being hashed.  If you want we can do a call to review the details of what is being hashed.

roconnor-blockstream:
@sanket1729 noted a problem that the _tapleaf_hash_ is not committed to in this scheme.

sanket1729:
ACK. Only reviewed the C code and convinced myself that the sighash algorithm commits to the necessary things to make sure that signatures are non-malleable.

Found some inconsistencies in the doc and code:
```
1) The description of `input-outpoints-hash` in the doc does not have the pegin information. The code has information about how pegins are dealt
2) `issuance-asset-amounts-hash` and `issuance-token-amounts-hash` have only the byte `0x00` in the doc. The code serializes `0x00` two times.
```

apoelstra:
ACK a1c61e2052cb75ac5de3b6bd0922817d43fccdbf except for the discrepancies that Sanket noticed.

roconnor-blockstream:
I've updated the documentation to address @sanket1729 's review.

apoelstra:
There is one more doubled-`00` to update for: in `issuance-token-amounts-hash`

But out of curiosity, why are these bytes doubled?

roconnor-blockstream:
They are doubled because there are ostensibly two fields, one for the asset and one for the amount.

apoelstra (approved):
ACK 4b053b922c30d1ebb583edc22ed5d1c82cbc9c02

I also reviewed only the C code, not the Haskell stuff. I also largely skimmed the boilerplate code, focusing on the actual sighash implementation. There I checked it against the text in the tech report, starting from the bottom in the hopes that I'd start to get tired/lazy at a different point than Sanket did.

PR #104, opened 2022-08-26
add GenPrimitives to cabal file

apoelstra:
This allows you to just do `cabal run GenPrimitives` and then `mv *.inc ./C/primitive/elements/`

### DIFF
--- a/Simplicity.cabal
+++ b/Simplicity.cabal
@@ -231,3 +231,29 @@ Test-Suite testsuite
                        tasty-quickcheck >=0.8 && <0.11,
                        transformers >=0.5 && <0.6,
                        vector >=0.12 && <0.13
+
+executable GenPrimitive
+    main-is:          GenPrimitive.hs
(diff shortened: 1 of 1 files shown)


roconnor-blockstream:
How do you feel about adding GenTests?

apoelstra:
I'm happy to do that, though I'm not sure how -- should it be added as another `Test-Suite` section? What does it do (it looks like the `main` function does writes the include files)?

roconnor-blockstream:
It would be exactly the same as GenPrimitives.  GenTests creates the `checkSigHashAllTx1.*` files and some other files.

apoelstra:
Ah, I misunderstood! I thought that `GenTests` somehow was a test for `GenPrimitives`.

Cool, I'll add it once I'm done doing a test run on my system.

roconnor-blockstream:
Merged in https://github.com/ElementsProject/simplicity/commit/119c5e9ec0465fe38cc11260a96871e6230a7aef.

PR #106, opened 2022-09-16
Generate Rust Jet modules

roconnor-blockstream:


roconnor-blockstream:
@apoelstra You don't need to review the code itself; however you should review the output.

apoelstra:
cc @chlewe 

I suggest we merge this as-is and then revisit it. It looks complete/correct as far as what it is, but it's incomplete -- in particular the `impl TryFrom<ElementsJetName> for &'static JetNode<Core>` block and similar also need to be autogenerated. Probably it'll take a bit of iteration to generate everything that we need to, and it might be better for you to do that based on this PR?

FYI with this PR you can run `cabal run GenRustJets` and it'll generate a bunch of `.rs` files in whatever directory you're in. Then you move those into `src/jet/init` in the Rust project and try to compile. Right now you get a ton of compilation errors because of references to the old names, but many of these look like code that should be autogenerated as well.

chlewe:
I'm also in favour of merging as is. In the future, we should reduce the number of required manual edits in Rust files.

I was thinking about merging the `*JetName` enum and `JetNode<*>` struct, but it is unclear how a Rust enum can contain the same data (CMR, types) for all enum variants. Maybe a `JetNodeInner` struct that is contained in all variants... The merge would make the `match` statements and `TryFrom` implementations clearer. The latter, we should automatically generate as part of the `init` files.

There was this idea of removing `TypeName`, but I do not see how we can represent types (as opposed to their names) in a const / static way. Rust is very restrictive here. We should add more names, like "byte", to shorten the generated names.

apoelstra (approved):
ACK 50ba60653a53d692574cdf58246558357dd9f0be

chlewe:
A range of jets seems to be missing from the generated files: `eq256`, `lt32`, `*_verify`, `sighash_all`, ...  Without these, the Miniscript compiler and basically all examples in rust-simplicity will not work. Are they missing because this is the first iteration, or has the way we use jets changed?

By the way, it would be cool if we had jets that handle bit strings of length `2^(2^n)` for arbitrary `n ≥ 0`, like in the jet catalogue. Not sure if this is possible.

roconnor-blockstream:
`SigAllHash` is there, but I'm missing a wrapper to build a standard signature scheme.

The other jets simply haven't been implemented as of yet.

roconnor-blockstream:
I've rebased this PR on some work in progress that adds jets for `CheckSig`, `Eq256`, `Le32`, and `Verify`.  Perhaps we will add  `EqVerify` and `LeVerify` jets down the road, but for now you will have to compose with the `Verify` jet.

As for the interpretation of Miniscript's checksig expression, as we should follow the [`checkSigVerify'` construction](https://github.com/ElementsProject/simplicity/blob/7b31f245fea74299fda00430ba08da30a751f160/Haskell/Core/Simplicity/Programs/CheckSig.hs#L59-L62) which is a combination of `disconnect` and a new `checkSigVerify` jet.

I've moved this PR into a draft state until the PRs this builds upto are completed and merged.  Until then, you can still use this branch for testing and experimentation.

roconnor-blockstream:
Okay, this is now rebased on master.  It is my understanding that more Rust code needs to be generated, so this isn't ready for merging yet.  However, there should now be more jets available.

uncomputable:
So, the `verify` jet is a shorthand for the `assert` construction? [What is `iden` doing in its definition?](https://github.com/ElementsProject/simplicity/blob/master/Haskell/Core/Simplicity/CoreJets.hs#L155) I would look this up myself, but I lack a Haskell IDE that allows me to jump to its definition _(suggestions appreciated)_.

uncomputable:
What is the value of `CheckSig` if all we do is `assert` / `verify` that its output is true? Is it a temporary jet until we have decided on how we bring `checkSigVerify'` into Rust?

uncomputable:
Speaking of `checkSigVerify'`, is it such a good idea to make use of `disconnect` in the standard conversion from Miniscript to Simplicity? Before universal sighashes make it into Bitcoin, this should be an optional add-on, I think. `disconnect` adds complexity that is currently not needed and it forces users to think about potential Bitcoin features that do not yet exist _(regarding how we teach this)_.

apoelstra:
@uncomputable I definitely would like universal sighashes in Elements. I think this is a super cool benefit of using Simplicity, which doesn't even require changing your upfront program, you just convert from Script to Simplicity and you're done.

Yes, there are open API/workflow questions about how to actually *use* universal sighashes but we can punt on that for now ... and if we instead implement some restricted form of sighash, then later we'll have to support two different `pk` equivalents in our simplicity-miniscript mapping.

roconnor-blockstream:
[`Verify`](https://github.com/ElementsProject/simplicity/blob/master/Haskell/Core/Simplicity/CoreJets.hs#L53) has type `Bit` to `()` and `Bit := () + ()`. The [`assert`macro](https://github.com/ElementsProject/simplicity/blob/2b29e59bccfc8f2682326306cc4b86ce51134268/Haskell/Core/Simplicity/Programs/Bit.hs#L50) process the right hand branch bringing the input type to `()`.  Then the `iden : () |- ()` expression passes that unit from the input to the output.  `unit : () |- ()` would also work just fine here.

---

In order to support batch verification of signatures, we only support `CheckSig` jets with verification semantics.  Tapscript does something similar.

---

I do think it is valuable to make `disconnect` part of the standard Miniscript translation.  Using `disconnect` means that UTXOs can immediately be used in covenants that require very fancy sighashes to operate. it is one of the best reasons to use Simplicity.  There are no real dangers of doing so.   The disconnected sighash programs must be signed, and so wallets shouldn't end up with any surprises.   I certainly think it is fine if wallets choose to only sign with a fix set of known programs.

uncomputable:
@roconnor-blockstream From what I understand, `Verify` is equivalent to `assert iden`. It is to be used as `t; Verify`. A more natural construction would be `assert t`, but this would be larger in program size. We cannot define `Verify` as `assert t`, because `t` cannot be jetted. This is why we resort to `assert iden`.

roconnor-blockstream:
One advantage of `assert t` is that it can leave it ambiguous as to whether there is a pruned branch there or not, but it comes at a substantial cost. `assert`s require a hidden node which is more than 32 bytes in size and maybe users aren't willing to pay that for branches that are pretty clearly actual asserts and not just a consequence of pruning.

roconnor-blockstream:
I've pushed a big update.  The encode and decode functions are still needed.  Probably other changes are needed too.  I'll revert to Draft for now, but feel free to leave comments.

Haskell-Generate/GenRustJets.hs
```diff
@@ -0,0 +1,346 @@
...
+rustFFIDoc :: Module -> SimpleDocStream a
+rustFFIDoc mod = layoutPretty layoutOptions $ vsep (map (<> line)
+  [ rustHeader
+  , rustFFIImports
+  , rustFFISigs mod
+  ])
+
+rustWrapperImports :: Doc a
+rustWrapperImports = vsep (map (<> semi)
+  [ "use super::{frame_ffi::CFrameItem, jets_ffi, CTxEnv}"
```
sanket1729: For some reason, this is generating two semicolons. 
`use super::{frame_ffi::CFrameItem, jets_ffi};;`

Haskell-Generate/GenRustJets.hs
```diff
@@ -0,0 +1,346 @@
...
+ where
+  wrapper (SomeArrow jet) = vsep
+   [ nest 4 $ vsep
+     [ pretty $ "pub fn "++cJetName jet++"(dst: &mut CFrameItem, src: CFrameItem, "++envParam++") -> bool {"
+     , pretty $ "unsafe { "++lowerRustModuleName mod++"_ffi::"++cJetName jet++"(dst, src, "++envArg++") }"
+     ]
+   , "}"
+   ]
+   where
+    envParam | CoreModule <- jetModule jet = "_env: const &CTxEnv"
```
sanket1729: Here and below: There should not be any const

Haskell-Generate/GenRustJets.hs
```diff
@@ -0,0 +1,346 @@
...
+  [ "use super::{frame_ffi::CFrameItem, jets_ffi, CTxEnv}"
+  ])
+
+rustWrappers :: Module -> Doc a
+rustWrappers mod = vsep ((<> line) . wrapper <$> moduleJets mod)
+ where
+  wrapper (SomeArrow jet) = vsep
+   [ nest 4 $ vsep
+     [ pretty $ "pub fn "++cJetName jet++"(dst: &mut CFrameItem, src: CFrameItem, "++envParam++") -> bool {"
+     , pretty $ "unsafe { "++lowerRustModuleName mod++"_ffi::"++cJetName jet++"(dst, src, "++envArg++") }"
```
sanket1729, line 331: should be `jets_ffi::` instead of `elements_ffi::`

Haskell-Generate/GenRustJets.hs
```diff
@@ -0,0 +1,346 @@
...
+rustImports mod = vsep (map (<> semi)
+  [ "use crate::bititer::BitIter"
+  , "use crate::bitwriter::BitWriter"
+  -- , "use crate::jet::elements::ElementsEnv"
+  , "use crate::jet::type_name::TypeName"
+  , "use crate::jet::Jet"
+  , "use crate::merkle::cmr::Cmr"
+  , "use crate::{decode_bits, Error}"
+  , "use bitcoin_hashes::sha256::Midstate"
+  , "use simplicity_sys::c_jets::CTxEnv"
```
sanket1729: This line is no longer needed

Haskell-Generate/GenRustJets.hs
```diff
@@ -0,0 +1,346 @@
...
+    , mempty
+    , "Cmr(Midstate(bytes))"
+    ]))
+  , "}"
+  ]
+ where
+  modname = rustModuleName mod
+
+rustJetTy fname getTy mod = vsep $
+  [ nest 4 (vsep (pretty ("fn "++fname++"(&self) -> TypeName {") :
```
sanket1729, line 176: These should not  have `TypeName` around it

```diff
- Core::Add32 => TypeName(b"*2i"),
+ Core::Add32 => b"*2i"
```

Haskell-Generate/GenRustJets.hs
```diff
@@ -0,0 +1,346 @@
...
+rustJetTargetTy :: Module -> Doc a
+rustJetTargetTy = rustJetTy "target_ty" (\(SomeArrow jet) -> unreflect (snd (reifyArrow jet)))
+
+rustJetPtr :: Module -> Doc a
+rustJetPtr mod = vsep $
+  [ nest 4 (vsep ("fn c_jet_ptr(&self) -> &'static dyn Fn(&mut CFrameItem, CFrameItem, &CTxEnv) -> bool {" :
+    [ nest 4 (vsep ("match self {" :
+        map (<>comma)
+        [ pretty modname <> "::" <> pretty (jetName jet) <+> "=>" <+>
+          pretty ("&simplicity_sys::c_jets::jet_wrapper::"++cJetName jet)
```
sanket1729, line 204: jet_wrapper should jets_wrapper

sanket1729 (commented):
Could not check elements.rs file because it generates all the jets. Left comments about `jets_wrapper.rs`, `core.rs`. `jets_ffi.rs` seems to be fine

Haskell-Generate/GenRustJets.hs
sanket1729: Given the application specific nature of core jets, it makes sense to have `elements_ffi` instead of `jets_ffi`. Will update rust code to reflect that

Haskell-Generate/GenRustJets.hs
roconnor-blockstream: done.

Haskell-Generate/GenRustJets.hs
roconnor-blockstream: done.

Haskell-Generate/GenRustJets.hs
roconnor-blockstream: Done.

Haskell-Generate/GenRustJets.hs
roconnor-blockstream: Done.

Haskell-Generate/GenRustJets.hs
roconnor-blockstream: Done.

sanket1729:
Files core.rs need import
```
use simplicity_sys::CFrameItem;
```
File elements.rs needs import 
```
use simplicity_sys::{CFrameItem, CElementsTxEnv};
use crate::jet::elements::ElementsEnv;
```

sanket1729:
Apart from the above changes, everything is good.

Haskell-Generate/GenRustJets.hs
```diff
@@ -271,16 +271,23 @@ rustHeader = "/* This file has been automatically generated. */"
...
-  [ "use crate::bititer::BitIter"
+  ([ "use crate::bititer::BitIter"
   , "use crate::bitwriter::BitWriter"
-  -- , "use crate::jet::elements::ElementsEnv"
   , "use crate::jet::type_name::TypeName"
   , "use crate::jet::Jet"
   , "use crate::merkle::cmr::Cmr"
   , "use crate::{decode_bits, Error}"
   , "use bitcoin_hashes::sha256::Midstate"
+  , "use simpicity_sys::CFrameItem"
```
sanket1729, line 284: /s/simpicity/simplicity

sanket1729 (commented):
One last spelling error. Everything else if good.

Haskell-Generate/GenRustJets.hs
roconnor-blockstream: Done.

Haskell-Generate/GenRustJets.hs
```diff
@@ -0,0 +1,377 @@
...
+rustFFIDoc mod = layoutPretty layoutOptions $ vsep (map (<> line)
+  [ rustHeader
+  , rustFFIImports
+  , rustFFISigs mod
+  ])
+
+rustWrapperImports :: Doc a
+rustWrapperImports = vsep (map (<> semi)
+  [ "use crate::CElementsTxEnv"
+  , "use super::{frame_ffi::CFrameItem, jets_ffi}"
```
sanket1729, line 331: Promise this is the last functional thing. 
```diff
- "use super::{frame_ffi::CFrameItem, jets_ffi}"
+ "use super::{frame_ffi::CFrameItem, elements_ffi}"
```

sanket1729 (commented):
This is the last one for sure

Haskell-Generate/GenRustJets.hs
roconnor-blockstream: Done.

sanket1729:
ACK, the rust generated code is compiles and builds successfully.

sanket1729 (commented):
Sorry about the review again. Requesting some changes would allow us to have our encoding/decoding temporarily run using the manual implementation we already have. 

This is useful in testing the simplicity program we create in rust via the FFI interface.

Haskell-Generate/GenRustJets.hs
```diff
@@ -0,0 +1,377 @@
...
+rustHeader = "/* This file has been automatically generated. */"
+
+rustImports :: Module -> Doc a
+rustImports mod = vsep (map (<> semi)
+  ([ "use crate::bititer::BitIter"
+  , "use crate::bitwriter::BitWriter"
+  , "use crate::jet::type_name::TypeName"
+  , "use crate::jet::Jet"
+  , "use crate::merkle::cmr::Cmr"
+  , "use crate::{decode_bits, Error}"
```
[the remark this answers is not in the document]
roconnor-blockstream: Done.

Haskell-Generate/GenRustJets.hs
```diff
@@ -0,0 +1,377 @@
...
+         | otherwise = "()"
+    envArg | Just "Bitcoin" == moduleName mod = "_env"
+           | otherwise = "env"
+    envBody | Nothing == moduleName mod = "env"
+            | Just "Bitcoin" == moduleName mod = "unimplemented!(\"Unspecified CJetEnvironment for Bitcoin jets\")"
+            | otherwise = "env.c_tx_env()"
+
+  encodeStub = vsep
+    [ "fn encode<W: Write>(&self, w: &mut BitWriter<W>) -> std::io::Result<usize> {"
+    , "    todo!()"
```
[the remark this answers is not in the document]
roconnor-blockstream: Done.

uncomputable:
Two nits about the currently generated code:
1. `cargo fmt` changes the code. Either we fix this or we run `cargo fmt` after each generation.
  - import order should be alphanumeric
  - long strings (`target_ty`) or function calls (`c_jet_ptr`) should be wrapped
2. The `Jet` methods are not in the same [order as in the trait](https://github.com/ElementsProject/rust-simplicity/blob/master/src/jet/mod.rs#L84-L107). `cargo clippy` doesn't complain about this but it is shown in my IDE.

uncomputable:
`cargo clippy` also complains about unused imports and unused function arguments. @roconnor-blockstream I could push a fixup commit (somehow) if this doesn't destroy your workflow.

sanket1729:
@uncomputable, I think we should figure out a way to disable `rustfmt` formatting lints for auto-generated code. Are you testing the latest code? I don't see any unused args or unused imports warnings

uncomputable:
I was using an older version. The clippy warnings seem to be gone. As for rustfmt, the import and method order are easy to fix. The wrapped return values are annoying, because shorter values _must not_ be wrapped, so I agree that we should disable the formatter.

Haskell-Generate/GenRustJets.hs
```diff
@@ -0,0 +1,448 @@
...
+rustImports mod = vsep (map (<> semi)
+  ([ "use crate::bititer::BitIter"
+  , "use crate::bitwriter::BitWriter"
+  , "use crate::jet::type_name::TypeName"
+  , "use crate::jet::Jet"
+  , "use crate::merkle::cmr::Cmr"
+  , "use crate::decode_bits"
+  , "use bitcoin_hashes::sha256::Midstate"
+  , "use simplicity_sys::CFrameItem"
+  , "use std::io::Write"
```
sanket1729, line 356: Needs "use crate::Error"

sanket1729 (commented):
Just one missing piece in enc/decoding APIs

Haskell-Generate/GenRustJets.hs
roconnor-blockstream: Done.

sanket1729:
The output rust code is compiling correctly. Are there any other features lacking for this PR?

roconnor-blockstream:
The Haskell pseudo-PR at the base of this PR needs tidying up.  I actually forgot about this, so I'll get on it.

roconnor-blockstream:
Rebased and squshed.

roconnor-blockstream:
Generated output is identical to 1d03dc9624ff242802d28bfdf7fb5562d40b71d4.

PR #108, opened 2022-10-28
Number jets

roconnor-blockstream:
This adds special "dynamically calculated" jets of type `1 ⊢ 2^(2ⁱ)` for word constants.

The serialization format is `0b10` followed by an encoding of `i` followed by 2^(i-1) bits.

This PR is solidifying the formal limits of the C implementation at 2^31.  The largest encoded number values must be strictly less than 2^31.  The maximum witness data blob size is 2^31-1 bits.  The maximum constant word jet size is 2^31 bits.  While in theory these would be "consensus critical limits", in Bitcoin and Elements, values are all limited to MAX_SIZE which is 2^20 bytes, far less than the limits being imposed by this library.  That is to say any Simplicity redemption that fits within the limits imposed by Bitcoin will not be able to reach these limits imposed by this C implementation of Simplicity.


apoelstra (approved):
ACK a037124b7b9f9e32f377e286ff288bf8a969881b

apoelstra:
Sorry for the long delay. I needed to reboot my development box and had to travel home over Christmas to do so.

apoelstra (approved):
ACK 8c3d94fb4dfbdd5f9dde8e7f3545b44fa2ef02c4

roconnor-blockstream:
@apoelstra I've just made some whitespace changes for you to review.

apoelstra (approved):
ACK 7429b30be15b330e437dc452c01d0531e3f95d96

PR #115, opened 2023-02-02
C eval case

roconnor-blockstream:
Adds some required anti-DOS checks.


C/eval.c
```diff
@@ -306,17 +306,42 @@ typedef struct evalState {
...
   size_t return_to;
-  union {
-    bool tcoOn;
-    bool bit; /* Scratch space used by the 'CASE' combinator. */
-  };
+  flags_type flags;
 } call;
 
+#define TCO_FLAG        0x01 // Whether TCO is on (1) or off (0).
+#define CASE_LAST_FLAG  0x02 // For case combinators, what was the last branch executed was right (1) or left (0).
```
apoelstra, line 319: In ed837d59f01fe25cd327d55b5753d09051e23f4c:

nit: `what was the` should be removed

C/eval.c
```diff
@@ -587,6 +597,35 @@ static bool runTCO(evalState state, call* stack, const dag_node* dag, type* type
...
   assert(pc == len);
+
+  return true;
+}
+
+/* Inspects the stack contents after a successful runTCO execution to verify anti-DOS properties:
+ * 1. All dag nodes were executed at least once.
+ * 2. Both branches of every CASE node were executed.
+ *
+ * If these are violoated it means that the dag should have had unused nodes prune or should have transformed
```
apoelstra, line 608: In 2d58ec81f4a848b7f7083e3e85ba85e78545b7b6:

nit `violated`

apoelstra: I would actually suggest rewriting this line, I find it hard to read. Maybe
```
If these are violated, it means that the dag had unused but unpruned nodes. For example, a CASE node may have two children, neither of which is an ASSERTL or ASSERTR node.
```

C/eval.c
```diff
@@ -587,6 +597,35 @@ static bool runTCO(evalState state, call* stack, const dag_node* dag, type* type
...
+ * Precondition: call stack[len];
+ *               dag_node dag[len];
+ */
+static bool antiDos(const call* stack, const dag_node* dag, size_t len) {
+  for(size_t i = 0; i < len; ++i) {
+    /* All combinators must be executed at least once. */
+    flags_type test_flags = EXEC_FLAG;
+
+    /* Both branches of every case combinator must be executed at least once. */
+    if (CASE == dag[i].tag) test_flags |= CASE_LEFT_FLAG || CASE_RIGHT_FLAG;
```
apoelstra, line 622: In 2d58ec81f4a848b7f7083e3e85ba85e78545b7b6:

Unless I'm crazy this code is just wrong -- it sets `test_flags` to the *logical or* rather than *bitwise or* of the two CASE_*_FLAGS. But then even if you use a bitwise or, the result is to set both flags on, and then the `if` below will check that *both* children were executed. But in the case that one of the children is an ASSERT, this check is incorrect.

C/eval.h
```diff
@@ -25,23 +30,25 @@
...
- * Precondition: NULL != evalSuccess
+ * Precondition: NULL != evalPass
  *               dag_node dag[len] and 'dag' is well-typed with 'type_dag' of type 1 |- 1;
  *               if 'dag[len]' represents a Simplicity expression with primitives then 'NULL != env';
  */
-static inline bool evalTCOProgram(bool *evalSuccess, const dag_node* dag, type* type_dag, size_t len, const txEnv* env) {
-  return evalTCOExpression(evalSuccess, NULL, 0, NULL, 0, dag, type_dag, len, env);
+static inline bool evalTCOProgram(bool *evalPass, const dag_node* dag, type* type_dag, size_t len, const txEnv* env) {
+  bool evalSuccess;
+  return evalTCOExpression(&evalSuccess, evalPass, NULL, 0, NULL, 0, dag, type_dag, len, env);
```
apoelstra, line 52: In 2d58ec81f4a848b7f7083e3e85ba85e78545b7b6:

I think it'd be clearer for `evalTCOExpression` to not take the `evalSuccess` argument at all, since it seems like you always ignore it.

C/eval.h
```diff
@@ -25,23 +30,25 @@
...
- * If during execution an 'assertr' or 'assertl' combinator fails, '*evalSuccess' is set to 'false'
- * Otherwise '*evalSuccess' is set to 'true'.
+ * If static analysis results determines the bound on memory allocation requirements exceed the allowed limits,
+ * set '*evalPass' to 'false'.
+ * If during execution an 'assertr' or 'assertl' combinator fails, '*evalPass' is set to 'false'.
+ * If after execution the anti-DOS checks fail, '*evalPass' is set to 'false'.
+ * Otherwise '*evalPass' is set to 'true'.
  *
- * Precondition: NULL != evalSuccess
+ * Precondition: NULL != evalPass
```
apoelstra, line 46: In 2d58ec81f4a848b7f7083e3e85ba85e78545b7b6:

I think this comment is out of date -- you can pass `NULL` for `evalPass` and it'll do the right thing (well, it will ignore success or failure entirely :)).

apoelstra:
Done review of 2d58ec81f4a848b7f7083e3e85ba85e78545b7b6. Looks great! Mostly just nits, except for the `antiDos` function which I don't understand, but fortunately is a very small part of the diff.

C/eval.c
roconnor-blockstream: You are not crazy.

C/eval.h
roconnor-blockstream: There is an API design here that I'm struggling with.  In some sense we should always fail when the anti-DOS tests fail.  On the other hand, we currently have tests (well currently a test) for sha256-compression that is unpruned, and if we wanted to test it on random input, it must be unpruned.

So my compromise is that the *program* evaluator (recall a program is an expression from unit to unit) always does the anti-DOS checks, while the lower level expression evaluator lets you optionally run or ignore the anti-DOS checks.

`evalTCOExpression`'s `evalSuccess` parameter is used in `test.c`. 

There is probably a million ways we could handle this success, check, pass, boolean logic, and I'm open to alternatives.  Currently `evalCheck` can only be true if `evalSuccess` is true, which is why `evalSuccess` is ignored in the above lines.

C/eval.c
roconnor-blockstream: Done.

C/eval.c
roconnor-blockstream: Replaced the text with just something simple.

C/eval.c
roconnor-blockstream: Fixed.

C/eval.h
roconnor-blockstream: Done.

C/eval.h
roconnor-blockstream: Obsolete.

apoelstra (approved):
ACK 825d57f5305f3c5b11680058c38de258d3797b0b

PR #117, opened 2023-02-15
Eliminate constant time fe_normalize.

roconnor-blockstream:
There is no nead for it to be in our call graph.


apoelstra (approved):
ACK d5d9730276e84e45eed897268d759fddba4133b3

PR #121, opened 2023-03-03
More careful support for threads.h

roconnor-blockstream:
`call_once` is slated for elimination by expanding our use of precomputed data.  However, in the meantime, we can make changes increase compiler support (in this case, for mingw32).


apoelstra (approved):
ACK efb4780de2a45ec99a5eb1039df8b0d24f54f47e

PR #123, opened 2023-03-10
Update to libsecp256k1 v0.3.0

roconnor-blockstream:
Relevant updates include

- new `secp256k1_fe_add_int`
- new `secp256k1_fe_half`
- new `secp256k1_fe_get_bounds` (only used for testing)
- new `secp256k1_fe_is_square_var`
- `secp256k1_ge_globalz_set_table_gej` updated to `secp256k1_ge_table_set_globalz`
- new `secp256k1_gej_eq_var` (currently unused)
- new `secp256k1_gej_cmov` (only for constant time algorithms)
- updated `secp256k1_gej_double` implementation.
- abstracted int128 implementation.

Many of these new functions ought to be allocated to jets.
Indeed, I believe a encoding has been set aside for `secp256k1_gej_eq_var`.

The `secp256k1_gej_double` algorithm has materially changed by returning a value whose `z` coordinate is half of what it used to be (and hence the `x` coordinate is ¼th of what it use to be and the `y` coordinate is ⅛th of what it used to be).  This requires us to update our simplicity specification to match.  It also has a knock-on effect on the `gej_add_ex` which needs to update it's z-ratio result in the case where it calls `gej_double` due to the change in double's z value.

Some other group operations have new implementations:

- `secp256k1_gej_add_var`
- `secp256k1_gej_add_ge_var`
- `secp256k1_gej_add_zinv_var`

However, these updates do not change the Jacobian coordinates of their selected representative.

There was an overhaul of `secp256k1_ecmult_odd_multiples_table` and related functions (including `secp256k1_ge_table_set_globalz`) in `ecmult` that I was involved with.  This also does not change the Jacobian coordinates of their selected representatives.

There are also a bunch of changes caused by the abstraction of the int128 data type.  Not only do these changes not change the computations done, they are involved in operations that have unique representatives (at least at the level of field element and scalar element abstraction that the Simplicity implementation operates at).


apoelstra (approved):
ACK 312841ac2b567aee201c9e39a80d235bba17e024

apoelstra:
Sorry for the delay. This was actually far easier to review than I expected.

PR #129, opened 2023-03-29
Consensus Memory Limits

roconnor-blockstream:
Add a dag length maximum value of 8 Mibi nodes.  The current encoding can fit at most two nodes per byte (two witness nodes).  As it stands, this dag length maximum value is unachievable given a blocksize limit of 4 Megabytes.  However, we may want to bring this limit down before final release.

We also add a cell limit of 5 Mibi cells.  This is slightly larger than Script's stack·item size limit in bits.

There is also a`NUMBER_OF_TYPENAMES_MAX` limit, but this is an implementation limit rather than a consensus limit.  Basically I didn't see a good way of exporting the `NumberOfTypeNames` value, so instead I put a bound of 1 Kibi on it for accounting purposes.  This bounds how large the various types of the jets are allowed to get.  It could be increased if it ever became necessary.  The current value of `NumberOfTypeNames` is 106.

### DIFF
--- a/C/dag.h
+++ b/C/dag.h
@@ -299,7 +299,7 @@ static inline size_t WITNESS_B(const dag_node* dag, const type* type_dag, size_t
  *
  * such that
  *
- *     0 < len
+ *     1 <= len <= DAG_LEN_MAX
  *
  * and for all 'i', 0 <= 'i' < 'len' and for all 'j', 0 <= 'j' < 'numChildren(dag[i].tag)',
  *
--- a/C/deserialize.c
+++ b/C/deserialize.c
@@ -3,6 +3,7 @@
 #include <assert.h>
 #include <limits.h>
 #include <stdlib.h>
+#include "limitations.h"
 #include "primitive.h"
 #include "unreachable.h"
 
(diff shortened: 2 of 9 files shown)


apoelstra:
All looks good. The last commit makes me a little nervous because it seems like the kind of assumption that might become wrong in the future but we'll forget to re-add the bounded math.

The 8 meg limit sounds good to me for purposes of this PR. Though it's not obvious to me what the actual memory usage will be, given these limits, because the limits are all in terms of Simplicity data structures rather than bytes.

roconnor-blockstream:
> All looks good. The last commit makes me a little nervous because it seems like the kind of assumption that might become wrong in the future but we'll forget to re-add the bounded math.

The fact `UWORDBound <= cellsBound` is pretty robust, and is all that is needed to ensure that if `cellBound` doesn't overflow than neither does `UWORDBound`.

> The 8 meg limit sounds good to me for purposes of this PR. Though it's not obvious to me what the actual memory usage will be, given these limits, because the limits are all in terms of Simplicity data structures rather than bytes.

For that information you will have to look at [the memory survey](https://github.com/BlockstreamResearch/simplicity/blob/memory/doc/memorySurvey.md) which is for 64 bit machines.  It's hard to say exactly the memory use because `sizeof(size_t)` varies from platform to platform (not to mention `sizeof(UWORD)`).  I'll rebase that survey on this PR as memory use has increased a little bit (though not peek memory use).

apoelstra (approved):
ACK 6f3a3e0bb96871284e3a381ed969baff1a5b44e3

PR #130, opened 2023-03-31
Add CPU cost limits.

roconnor-blockstream:
CPU costs counted in terms of weight units.   The benchmark timings are normalized so that the `CheckSigVerify` takes 100 weight units.  However, because `CheckSigVerify` is a batchable jet, it gets a 50% discount, giving it a total of a total of 50 weight units.

These benchmarks place `Sha256Block` at 1.046 weight units.

During static analysis, cost is accumulated in terms of milli-weight units before being compared to the total allotted budget in weight units.


C/eval.c
```diff
@@ -657,13 +658,17 @@ typedef struct memBound {
...
  *                     or 'dag_bound->extraCellsBound' characterizes the number of cells needed during evaluation of 'dag'
  *                        and 'dag_bound->extraUWORDBound' characterizes the number of UWORDs needed
  *                              for the frames allocated during evaluation of 'dag'
- *                        and 'dag_bound->extraFrameBound[0]' bounds the the number of stack frames needed during execution of 'dag';
+ *                        and 'dag_bound->extraFrameBound[0]' bounds the the number of stack frames needed during execution of 'dag'.
+ *
+ * :TODO: The cost calculations below are estimated and need to be replaced by experimental data.
  */
 static bool computeEvalTCOBound(memBound *dag_bound, const dag_node* dag, const type* type_dag, const size_t len) {
+  const ubounded overhead = 10;
```
apoelstra, line 671: In 57f9a933a48fcfd2a9c039a38e33f55cbcfa807c:

I'm a bit confused about this value. It looks like the cost of the 9 core combinators (and witness and disconnect). But aren't these already priced-in as ~half a byte by the fact that they're explicitly encoded in the program in the tx witness?

Also, 10 seems really really high. I guess that the units here are witness bytes, judging by the normalization to "checksig == 50" in the previous commit.

C/eval.c
```diff
@@ -833,26 +833,30 @@ static bool computeEvalTCOBound(memBound *dag_bound, const dag_node* dag, const
...
+ *               budget <= BUDGET_MAX
  *               if 'dag[len]' represents a Simplicity expression with primitives then 'NULL != env';
  */
 bool evalTCOExpression( bool *evalSuccess, flags_type anti_dos_checks, UWORD* output, ubounded outputSize, const UWORD* input, ubounded inputSize
-                      , const dag_node* dag, type* type_dag, size_t len, const txEnv* env
+                      , const dag_node* dag, type* type_dag, size_t len, ubounded budget, const txEnv* env
                       ) {
   assert(1 <= len);
   assert(len <= DAG_LEN_MAX);
+  if (BUDGET_MAX < budget) budget = BUDGET_MAX; /* Soft enforcement of budget precondition. */
```
apoelstra, line 844: In a4a3ed3b0eee75521640a2c7fefc0e09127a77b5:

I think this should just be an assertion.

C/eval.c
apoelstra: Ah, never mind, in the first commit we define the `Weight` type to be in milliweight (and also you say this in the PR description). So this is actually 0.01 weight. Also there is a comment in the Haskell saying "this is just an estimate".

Ok, that answers all my questions.

Though I kinda wish that this hardcoded 10 came as a `#define` in one of the .inc files, to ensure that the Haskell and C are consistent. Fine for now.

apoelstra (approved):
ACK a4a3ed3b0eee75521640a2c7fefc0e09127a77b5

C/eval.c
roconnor-blockstream: done.

PR #131, opened 2023-04-12
Anti-DOS for Hidden nodes

roconnor-blockstream:
Anti-DOS should not expect Hidden nodes to be executed. Without this fix, programs that require pruning cannot pass the Anti-DOS checks.

Test cases will be in a forthcoming pruning PR.


apoelstra (approved):
ACK 987bf774940ffd1f6c5ade4bb47568158f968c23

PR #133, opened 2023-04-14
Replace fails with assertions

roconnor-blockstream:
Since fail cannot be deserialized (in the C implementation), it is better to use assert. Because these functions are jetted in practice, this only really makes a difference when testing unjetted code.


roconnor-blockstream:
:warning: This will change the CMR of several jets and will need to be synced with Rust.

apoelstra (approved):
ACK 53101b15465fae093592d2d58fed66ef75eff456

PR #134, opened 2023-04-14
Allow null program

roconnor-blockstream:
Programs can be null, so long as their size is 0.
Of course, an empty program will eventually fail, but it shouldn't be an illegal invocation.

### DIFF
--- a/C/primitive/elements/exec.c
+++ b/C/primitive/elements/exec.c
@@ -35,7 +35,8 @@ extern bool elements_simplicity_execSimplicity( bool* success, unsigned char* im
                                               , int64_t budget
                                               , const unsigned char* amr
                                               , const unsigned char* program, size_t program_len) {
-  if (!success || !tx || !taproot || !program) return false;
+  if (!success || !tx || !taproot) return false;
+  assert(NULL != program || 0 == program_len);
 
   bool result;
   combinator_counters census;


apoelstra:
LGTM. It looks like if you pass a NULL `program` with a 0 length then `decodeMallocDag` will fail cleanly. Though without this assertation, I believe passing a null pointer with a nonzero length will cause a null dereference. It may be worthwhile adding an extra assert to catch this.

apoelstra (approved):
ACK 008fd8bc662f4c204db3db71e986c7e17bd568fd

roconnor-blockstream:
Added an assert.

uncomputable:
ACK [c0d893c](https://github.com/BlockstreamResearch/simplicity/pull/134/commits/c0d893c8007629982119fc01bec8bdb902572b21). Looks good, since the preconditions exclude the case that other variables are null.

I assume that the precondition `unsigned char program[program_len]` means that when `program` is null, `program_len` must be zero.

uncomputable (approved):

roconnor-blockstream:
> ACK [c0d893c](https://github.com/BlockstreamResearch/simplicity/commit/c0d893c8007629982119fc01bec8bdb902572b21). Looks good, since the preconditions exclude the case that other variables are null.
> 
> I assume that the precondition `unsigned char program[program_len]` means that when `program` is null, `program_len` must be zero.

I've been a bit vague on this issue.  My current understanding is that officially the C standard says that the NULL pointer cannot be a zero length array.  On the other hand, I think it is a distinction without an observable different, so I figure I can allow it.

apoelstra:
>My current understanding is that officially the C standard says that the NULL pointer cannot be a zero length array.

I believe this is correct, but because C++ STL objects seem to do this (well, they have "empty" forms of various objects that contain NULL pointers, which when interfacing with C via `.data()` and `.size()` appear as 0-length null-pointer arrays), we probably want to explicitly handle it.

apoelstra (approved):
ACK c0d893c8007629982119fc01bec8bdb902572b21

PR #136, opened 2023-04-20
Add coverage analysis

roconnor-blockstream:


apoelstra (approved):
ACK 4168bed41e179b134220923b74b6b234394055cd

Simplicity.C.nix
```diff
@@ -1,13 +1,22 @@
-{ lib, stdenv, wideMultiply ? null }:
+{ lib, stdenv, gcovr ? null, wideMultiply ? null, withCoverage ? false
+, gcov-executable ? "gcov" # Set to "${nixpkgs.libllvm}/bin/llvm-cov gcov" when using clang.
```
apoelstra, line 2: I'd suggest wrapping `gcovr` and `gcov-executable` into a set (though have the whole thing default to `null`) and then passing either a set or `null` in.

Then in default.nix the user can pass either a set or `null`, and it will be forwarded.

I've been able to run this just by passing `withCoverage` as `true` into default.nix, and confirmed that it produces some sort of coverage output (though I didn't look at it), so I"m kinda confused about what these extra variables are...

Simplicity.C.nix
```diff
@@ -1,13 +1,22 @@
...
+}:
 assert wideMultiply == null
     || wideMultiply == "int64"
     || wideMultiply == "int128"
     || wideMultiply == "int128_struct";
+assert withCoverage -> gcovr != null;
```
apoelstra, line 8: ...and I'm confused as to how I avoided triggering this assertion in my code.

apoelstra:
I ACKed since I was able to use the `withCoverage` parameter and it produced some coverage output, both with gcc and clang, but I am confused about the details.

Simplicity.C.nix
roconnor-blockstream: gcovr is part part of nixpkgs and so it is automatically included.

Maybe I'll just go ahead and delete this assertion.  It is a bit silly.

Simplicity.C.nix
roconnor-blockstream: I've reworked this line.

roconnor-blockstream:
I've reworked the logic so now everything works with `clangStdenv` without requiring any thought.

roconnor-blockstream:
I've added back the assertion.

The explanation is that `gcovr` is part of nixpkgs, so it is always filled in, but simplicity doesn't need `gcovr` to be an argument if you are not doing coverage analysis. So if you are compiling on an exotic system whose packages do not contain `gcovr` then it is still fine so long as you don't turn on `withCoverage`.

apoelstra (approved):
ACK 5a13124d0bd6b8506e5a8f2859cb9d6c99c0ae4e

Issue #12, opened 2019-08-01
Keep assertions in production builds of C library


roconnor-blockstream:
It is preferable to keep assertions (and aborting on UNREACHABLE) on even in production builds, with the possible exception of important code paths (such as the main eval loop and jets).  It is maybe acceptable to add an OPTIMIZED flag (off by default) to turn off these safety checks for those who wish to risk it.

PR #137, opened 2023-05-02
Simplicity assertions

roconnor-blockstream:
Refactor all assertions through this simplicity_assert.h module.

`simplicity_debug_assert` is used within the main evaluation loop, and those assertions will be removed if `PRODUCTION` is declared.

Like Bitcoin and Elements, `NDEBUG` is not allowed.  Use `PRODUCTION` instead.


roconnor-blockstream:
Fixes #12.

roconnor-blockstream:
`NDEBUG` in the makefile needs to be eliminated.

apoelstra (approved):
ACK 02ce7b6798434590a06f8a9b652bec012d709920

apoelstra (approved):
ACK d5cf57a69940bd96653944df1f58d5564afc68be

apoelstra:
This exposes `doCheck` from the C nix file but it can't be set from default.nix.

PR #140, opened 2023-05-09
Error codes

roconnor-blockstream:
Return error codes upon failure.


roconnor-blockstream:
@apoelstra Please give this a careful review.

C/errorCodes.h
```diff
@@ -15,7 +16,23 @@
...
+#define SIMPLICITY_ERR_AMR (-32)
+#define SIMPLICITY_ERR_EXEC_BUDGET (-34)
+#define SIMPLICITY_ERR_EXEC_MEMORY (-36)
+#define SIMPLICITY_ERR_EXEC_JET (-38)
+#define SIMPLICITY_ERR_EXEC_ASSERT (-40)
+#define SIMPLICITY_ERR_ANTIDOS (-42)
 
-#define PERMANENT_FAILURE(err) (!((err) & 1))
+/* Check if failure is permanent or success (which is always permanent). */
+#define PERMANENT(err) (!((err) & 1))
```
apoelstra: In 379a00037f08348746dc9dd3940600f4f798e610:

I'd prefer if this were called `IS_PERMANENT` to be clear that it's checking a condition rather than marking an existing error as permanent somehow.

apoelstra:
This can go in an additional commit, but I think it'd be better to do

```
typedef enum {
    SIMPLICITY_ERR_OK = 0,
    SIMPLICITY_ERR_MALLOC = -1,
    ...
} simplicity_error;
```
and then use `simplicity_error` in place of `int` everywhere you are handling error values (rather than integers). You probably also want an `#define OK(err) ((err) == SIMPLICITY_ERR_OK)` macro.  This will let you
* hopefully get warnings when comparing/assigning to literal integers, except probably the literal `0` because in C this gets coerced to literally every single non-struct type
* hopefully get warnings if you `switch` on an error and forget cases
* make it a bit easier for the reader of the code to tell what's an error, what's not

Sadly it will not actually enforce any additional type safety over using `int`, except for the warnings (which are not mandated by the C standard) that gcc and clang sometimes provide. AFAICT there is no way to get remotely type-safe errors in C except to define an error *struct*, and then you lose the `==` operator and cannot even use `memcmp` due to padding fears. In the Blockstream Liquid HSM we took this route, along with a bunch of macros for doing comparisons, logging-then-propagating errors, etc. I'm not sure it was worth the pain.

apoelstra:
Coercions from `int` are especially insidious because the C and POSIX stdlibs return ints all over the place and appear to have chosen whether 0 means "success" or "failure" independently randomly for each function. So if a programmer were foolishly to think that he could just propagate a return value without somehow manually converting it, his program would be wrong.

Done reviewing 22128590527dc3ac9a5492e56d3e28177a9c9dff. Looks good except for the notes that I posted, which are not really serious.

roconnor-blockstream:
Do you want an enum as part of the library ABI or just for internal use?  We do compile it as part of elements, so arguably it doesn't matter.

roconnor-blockstream:
Missing conversion for `decodeJet`.

C/deserialize.c
```diff
@@ -24,17 +24,17 @@ static int32_t getWord32Array(uint32_t* result, const size_t len, bitstream* str
     if (bits16 < 0) return bits16;
```
apoelstra, line 24: In 9d3240e4468533acd17e2d14fc1eae85edf42ef4:

Should cast `bits16` here (and three lines above).

roconnor-blockstream: If only we had compiler warnings ...

apoelstra:
Done reviewing 012d5401ea2da842a70a1d82877281dc6f5b57fc.

Found one nit (missing cast). Not confident that this is the only one, but it's largely a stylistic thing.

Spent some time experimenting with a followup to this PR which wraps the enum in a struct like
```
typedef struct {
  simplicity_err_ty ty;  // this is the existing enum
  int res;
} simplicity_err;

#define ERR(e) ((simplicity_err) { .ty = e })
#define ERR_EQ(e1, e2) ((e1).ty == (e2).ty)
#define OK(e) ((simplicity_err) { .ty = SIMPLICITY_NO_ERROR; .res = e })
```

Then all your casts become ERR()s and it becomes possible to return an error type directly from `readNBits` and other "maybe returns an int, maybe returns an error" functions. I suppose you could also define a macro which extracts the `.res` field, but only after checking that `.ty == SIMPLICITY_NO_ERROR` (and otherwise asserting, or maybe doing an early return, depending whether you were comfortable hiding control flow inside of macros).

I made it maybe 5-10% through the project. It seemed like it would work, it'd just be tedious, and I didn't have a clear sense of progress because the compiler was only giving me errors from one file at a time for some reason. If we were to go this route we should do it in a followup PR.

apoelstra (approved):
ACK 012d5401ea2da842a70a1d82877281dc6f5b57fc

C/deserialize.c
roconnor-blockstream: Fixed.

apoelstra (approved):
ACK ed507e688482089a73250e6bf5a9bebc2164a84e

PR #144, opened 2023-05-16
Update precondition

### LINKED ISSUE #142: Some confusion about type inference for constant words
I have a couple questions about type inference for const words.

1. I believe the doccomment on `typeInference` should be updated: https://github.com/BlockstreamResearch/simplicity/blob/7be9a8725135ea81f6454968af574a6ca2390407/C/typeInference.c#L244-L245 Currently it says as a precondition that if a node is a `JET` then the `bound_var` variable must be correctly initialized; since adding `WORD`, the precondition should also be required if the node is `WORD`.

2. Why does `UNIT` leave its source type unbound while `WORD` fixes the source to be unit? It seems like `UNIT` could be thought of as a trivial form of `WORD` but this is a conspicuous difference. (I guess that unit is supposed to serve a "read and throw away an arbitrary input" role?)

roconnor-blockstream:
Fixes #142.

### DIFF
--- a/C/typeInference.c
+++ b/C/typeInference.c
@@ -243,6 +243,8 @@ static size_t max_extra_vars(const combinator_counters* census) {
  *               unification_var bound_var[N];
  *               forall '0 <= i < len', if 'dag[i].tag == JET'
  *                                      then 'dag[i].sourceIx < N' and 'dag[i].targetIx < N'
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK 2d45afe1ba392619700ff9c0e8c4f99cbe66697b

Issue #141, opened 2023-05-14
Should add nix target for building .pdf (and .tex?) version of the tech report


apoelstra:

roconnor-blockstream:
Related: https://savannah.gnu.org/bugs/?32171

roconnor-blockstream:
Let's keep this issue open even after #145.  When a new version of TeXmacs is released we can replace `xvfb-run` with just a simple `--headless` argument to TeXmacs.

roconnor-blockstream:
--headless will probably be available in Texmacs 2.1.4, which was merged into [nixpkgs on 2024-09-08](https://github.com/NixOS/nixpkgs/pull/340621)

We should wait for this to get into nixos-stable

PR #145, opened 2023-05-17
Target for generating PDF

roconnor-blockstream:
Texmacs does not make this easy.

Related to #141.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -42,6 +42,22 @@ let hp = nixpkgs.haskell.packages.${ghc};
     ccomp-platform = "x86_64-linux";
   };
 
+  pdf = nixpkgs.runCommand "Simplicity-TR" {} ''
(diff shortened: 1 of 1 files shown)


apoelstra:
I have a mild preference for using `mxDerivation` so that I have the `overrideDerivation` attribute, which I use to add extra metadata to my derivations. But this is fine.

apoelstra (approved):
ACK a365c82347a61b5300a9b02d6f6adb4b7bccaeae

roconnor-blockstream:
P.S.  It isn't `mkDerivation` that adds `overrideDerivation`, it is `callPackage` that does.  `runCommand` actually uses `mkDerivation`.  `mkDerivation` does add `overrideAttrs` which is the new-and-improved version of `overrideDerivation`.

PR #147, opened 2023-05-30
Add some Arith jets

roconnor-blockstream:
- Full Increment
- Increment
- Negate
- Full Decrement
- Decrement
- Is Zero
- Is One
- Lt
- Min
- Max
- Median


apoelstra:
Looks good. I'm a little bit wary of the wrapping overflow behaviour in increment/decrement, given the troubles the EVM has had with this (although we do output an overflow bit so it's easy to detect/assert on it). But given that "nobody should be directly writing Simplicity" I think this is fine. It's well-defined and obvious how one would constrain it to get safer behaviour, for various definitions of "safer".

`nix-build -A haskell` fails on this.

But otherwise 981b63b75d78fd3fd31b0fff8af656eb20a79c8a looks good.

apoelstra (approved):
ACK d1c49bb9e8053932b6191a8d772a569a3eb613fb

PR #148, opened 2023-06-01
Add some Arith jets

roconnor-blockstream:
- Div Mod
- Divide
- Modulo
- Divides


apoelstra:
I really don't think we should define division by 0 to be 0. Would be better to add a division-by-0 bit.

roconnor-blockstream:
When dividing `a` by `b`, the fundamental equation is that

   `b` × `q` = `a` - `r`.

In the case where `b = 0`, we have defined the remainder `r = a` (because modding Z out by the ideal generated by (0) yields an isomorphic copy of Z, and thus the equivalence class that `a` maps to only contains `a` itself. Thus `r` , the canonical member of this equivalence class, can only be `a` itself). This means that the fundamental equation is satisfied by any value for the quotient `q`. I've chosen `0` as a canonical choice, which is a common choice: https://coq.inria.fr/library/Coq.ZArith.BinIntDef.html#Z.div_eucl and https://math-comp.github.io/htmldoc/mathcomp.ssreflect.div.html#edivn

roconnor-blockstream:
Wrong ~~channel~~ PR.

PR #152, opened 2023-06-06
Fix 151

### LINKED ISSUE #151: Assertion failure in `decodeWitnessData`
`decodeWitnessData` calls `readBitstring`, which has an assertion on the length to read. But the length to read is an unchecked value read from the serialized program. So you can trigger an assertion.

`761f001a00`, whatever that is, seems to trigger this.

Found by fuzzer.

roconnor-blockstream:
Fixes #151.

### DIFF
--- a/C/test.c
+++ b/C/test.c
@@ -435,6 +435,20 @@ static void regression_tests(void) {
     const unsigned char regression1[] = {0x27, 0xe1, 0xe0, 0x00, 0x00, 0x00, 0x00};
     test_program("regression1", regression1, sizeof(regression1), SIMPLICITY_ERR_DATA_OUT_OF_RANGE, NULL, NULL, NULL);
   }
+  {
+    /* Unit program with incomplete witness of size 2^31-1. */
+    const unsigned char regression2[] = {0x27, 0xe1, 0xdf, 0xff, 0xff,  0xff, 0xff};
+    test_program("regression2", regression2, sizeof(regression2), SIMPLICITY_ERR_BITSTREAM_EOF, NULL, NULL, NULL);
+  }
+  {
+    /* word("2^23 zero bits") ; unit */
+    size_t sizeof_regression3 = ((size_t)1 << 20) + 4;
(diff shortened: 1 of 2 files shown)


roconnor-blockstream:
@apoelstra Do you prefer this PR squashed?

apoelstra:
Also, can you confirm that if somebody tries to make a witness of size 2^32 + 1, that the program is rejected without the assertion triggering? I believe the original bug occured in `decodeWitnessData` which doesn't understand types or anything, it just reads the total witness length and tries to cram it all into a bitstring.

apoelstra:
Finally, I am getting build failures like
```
       > test.c:432:13: error: function declaration isn't a prototype [-Werror=strict-prototypes]
       >   432 | static void regression_tests() {
       >       |             ^~~~~~~~~~~~~~~~
       > cc1: all warnings being treated as errors
```
I suspect because I have a newer gcc than you. I think you have to stick a `void` into the argument list for the function.

roconnor-blockstream:
Oops.  I was testing with clang.

roconnor-blockstream:
Completed all suggestions.

C/test.c
```diff
@@ -435,6 +435,20 @@ static void regression_tests(void) {
     const unsigned char regression1[] = {0x27, 0xe1, 0xe0, 0x00, 0x00, 0x00, 0x00};
     test_program("regression1", regression1, sizeof(regression1), SIMPLICITY_ERR_DATA_OUT_OF_RANGE, NULL, NULL, NULL);
   }
+  {
+    /* Unit program with incomplete witness of size 2^31-1. */
+    const unsigned char regression2[] = {0x27, 0xe1, 0xdf, 0xff, 0xff,  0xff, 0xff};
```
apoelstra, line 440: What if the incomplete witness has size 2^31 + 1?

apoelstra (approved):
ACK 8d2bdb426a55bff6c19ff8fd371d4abdd71f50b3

PR #153, opened 2023-06-07
Correct rustJetCmr

roconnor-blockstream:
Previously this function was creating the IMRs of jet specifications, not their CMRs.  This commit changes the behaviour to return the jet CMRs.


roconnor-blockstream:
@apoelstra @uncomputable This PR requires changes on the Rust side as it removes the need to compute jet CMRs from the IMR of their specifications.  The required changes  are analogous to what was done in the C code in #122.

apoelstra (approved):
ACK 51e8665dd620ed571b15528228547663eec8fffb

apoelstra:
Thanks for the heads up! I didn't actually test the generated code, but the diff looks fine.

PR #154, opened 2023-06-07
Add some Word jets

roconnor-blockstream:
- High
- Complement
- And
- Or
- Xor
- Maj
- Xor3
- Ch
- Some
- All


C/jets.c
```diff
@@ -34,6 +34,145 @@ LOW_(16)
...
+HIGH_(16)
+HIGH_(32)
+HIGH_(64)
+
+#define COMPLEMENT_(bits)                                                 \
+/* complement_n : TWO^n |- TWO^n */                                       \
+bool complement_##bits(frameItem* dst, frameItem src, const txEnv* env) { \
+  (void) env; /* env is unused. */                                        \
+  uint_fast##bits##_t x = read##bits(&src);                               \
+  write##bits(dst, ~(1U*x));                                              \
```
apoelstra, line 55: In ac9fc58de9ebe2527f594b31381943d387340fc5:

Is `1U*` actually needed here (or elsewhere it is used)? On first reading I thought it was there to promote signed integers to unsigned ones, but there are actually no signed integers anywhere here. So I guess it is for some other promotion-related reason.

roconnor-blockstream: Things like `uint_fast_8_t` are promoted to signed integers before bitwise complement is applied to it.  And I think you would agree that taking a bitwise complement of a signed integer is sketchy.  It certainly could tickle trap representations on obscure C implementations.

I used to have these unsigned integer promotions everywhere, but I eventually dropped them for `&`, `|`, and ``^`.  Although doing bitwise operations on signed integers is still sketchy, they are at least on signed integer values that are from promoted unsigned values, and, since the sign bit is never set, not much bad can happen with those bitwise operations.

apoelstra: > Things like `uint_fast_8_t` are promoted to signed integers before bitwise complement is applied to it.

Lol what a shitty language.

I don't fully buy that "the sign bit is never set" by `|` or `^`. I believe they can be, though only in the case that the width of the underlying type is the same as the width of the signed integer. And in this case, I think we are still fine, because the unsigned->signed->unsigned conversion will preserve the value rather than doing sign extension anywhere.

apoelstra:
The C code does not build after the second commit.

C/jets.c
roconnor-blockstream: An unsigned value that is promoted to a signed integer never has the sign bit set because of the promotion requirement that every (positive) value fits inside the range of the signed integer.

Since none of the promoted values have their sign bit set, the  `&`, `|`, and `^` operations never set the sign bit either.

apoelstra: Ok, thanks, I understand now.

roconnor-blockstream:
Fixed.

apoelstra (approved):
ACK 3925d3f4790d82d7ef0143703b0d77dde3b077e8

PR #156, opened 2023-06-15
Early failure for HIDDEN root nodes.

### LINKED ISSUE #155: A program which is just a hidden node shouldn't decode
Example: `300000000000000000000000000000000000000000000000000000000000000000`

I would argue that a sole hidden node is a "hidden node outside of a CASE combinator" and should fail to decode with an `SIMPLICITY_ERR_HIDDEN`.

I can see both sides of this argument philosophically, but pragmatically, I'd also prefer that this doesn't decode, because in rust-simplicity we're planning to drop hidden nodes entirely, which means *we* won't be able to decode it, and I'd like to be able to fuzz that the C and Rust implementations decode exactly the same set of programs. (Well, except that the Rust code will decode fail nodes.)

roconnor-blockstream:
Currently if a DAG consists of exactly one node that is a HIDDEN node it passes all validation until finally failing during evaluation. This commit causes this case to fail right away.

Fixes #155


apoelstra (approved):
ACK ae358ab2a682a48765956d5935ffaafd781af21a

apoelstra:
Confirmed that this now matches the Rust code in this case.

PR #158, opened 2023-06-21
Update imports in autogenerated Rust jets

uncomputable:
`Bit{Iter,Writer}` are exported at the crate root.


apoelstra:
We also need to change `fn decode` to return a `decode::Error` rather than an `Error`.

uncomputable:
Pushed a commit that updates the error types.

apoelstra:
This looks good. Checked that when I generate the Rust code from this branch, I can remove all the "FIXME"s related to symbol paths in the autogenerated code.

apoelstra (approved):
ACK e5f74ca15ed553b769f3862d68b72271d1a98bd0

PR #161, opened 2023-06-27
Add cost to autogenerated Rust jets

### LINKED ISSUE #160: Jet costs as part of Haskell library
The cost of a jet should be part of the Haskell library, [as said in this todo](https://github.com/BlockstreamResearch/simplicity/blob/e5f74ca15ed553b769f3862d68b72271d1a98bd0/Haskell-Generate/GenPrimitive.hs#L175). This would make code generation for Rust much easier and lead to less code duplication.

uncomputable:
This adds a cost function to the Rust jets.

The Rust API is only 90% finished so I will make this a draft.

We can fix #160 before merging this if you want.


roconnor-blockstream:
While ultimately those jet benchmarks need to be properly incorporated into the main library, for now I'd like to move the bechmark data into it's own module that can be then imported into both the C and Rust generation code.

If you like I can prepare that for you.

uncomputable:
@roconnor-blockstream That would be great, thank you. Modules and imports are always a bit fiddly.

roconnor-blockstream:
See #163.

roconnor-blockstream:
#163 is now merged, so you can go ahead and rebase your work on it.

roconnor-blockstream:
You will want to update the `Simplicity.cabal` file like was done in https://github.com/BlockstreamResearch/simplicity/pull/163/files#diff-da7115fd39036c9029c55c740d8920bcbed7adc04ed85b6a673d1a05de0c0906

uncomputable:
Rebased and ready for review.

For some reason `cabal` built the executable without specifying the `other-modules`. Maybe this was due to my local state... I added the line just to be safe.

roconnor-blockstream:
Yes, it will still work, but emit a warning if you don't update the `.cabal` file.

Haskell-Generate/GenRustJets.hs
```diff
@@ -234,6 +236,24 @@ rustJetDecode mod =
...
 
+rustJetCost :: Module -> Doc a
+rustJetCost mod = vsep $
+  [ nest 4 (vsep ("fn cost(&self) -> Cost {" :
+    [ nest 4 (vsep ("match self {" :
+        map (<>comma)
+        [ pretty modname <> "::" <> pretty (jetName jet) <+> "=>" <+>
+          if BitcoinModule == jetModule jet
+          then "unimplemented!(\"Unspecified cost of Bitcoin jets\")"
+          else "Cost::from_jet(" <> (pretty $ milliWeight $ cost $ jetName jet) <> ")"
```
roconnor-blockstream, line 247: I believe the usual idiom in Haskell is to write

    pretty . milliWeight . cost $ jetName jet

uncomputable: Function composition, nice. I pushed an update.

Haskell-Generate/GenRustJets.hs
```diff
@@ -234,6 +236,24 @@ rustJetDecode mod =
...
                        ) <> line
 
+rustJetCost :: Module -> Doc a
+rustJetCost mod = vsep $
+  [ nest 4 (vsep ("fn cost(&self) -> Cost {" :
+    [ nest 4 (vsep ("match self {" :
+        map (<>comma)
+        [ pretty modname <> "::" <> pretty (jetName jet) <+> "=>" <+>
+          if BitcoinModule == jetModule jet
+          then "unimplemented!(\"Unspecified cost of Bitcoin jets\")"
```
apoelstra, line 246: In b4d9fcdea97eb293db03ccd8ea4f57071dc8b86c:

We should move the `unimplemented!` up to the entire function instead of doing the match and then repeating it a zillion times.

uncomputable: We do the same for Bitcoin jet pointers, but I can update both.

apoelstra: Updating both sounds good, yeah.

uncomputable: Done

apoelstra:
Looks good except I'd like you to rename `from_jet` to `from_milliweight`.

apoelstra (approved):
ACK 7e9cee39f3237b15c3513a83d2ef1deeb6c3714f

PR #162, opened 2023-06-27
Clarify conditons for bindings

roconnor-blockstream:
Bound unification variables should never be attempted to be rebound to themselves.  In particular when unifying a variable with itself, it should never try to apply bindings against itself.

However, if for whatever reason this does happen, and assertions are turned off, then simply proceed without creating new unification constrains (which would already hold anyways).


apoelstra:
I have this code fresh in my head and agree with this reasoning (and with the solution).

apoelstra (approved):
ACK 6825c2c32a629ffdbfb7a5b7d29379ce0d242533

PR #165, opened 2023-07-05
Incorporate jet Weight costs into the parameters of jets

roconnor-blockstream:
This is a first step towards completing #160.  However, it remains to actually incorporate the jet costs into their CMR specifications.


apoelstra (approved):
ACK 54b117211c53ffb1a217cea9c29d0a622364711a

Issue #159, opened 2023-06-26
Incorrect weight calculations for Witness and Disconnect


roconnor-blockstream:
@uncomputable noticed the Haskell implementation of the weight cost of terms for Witnesses appears incorrect.

https://github.com/BlockstreamResearch/simplicity/blob/e5f74ca15ed553b769f3862d68b72271d1a98bd0/Haskell/Core/Simplicity/BitMachine/StaticAnalysis/Cost.hs#L40

It should have overhead and use `milli`.

uncomputable:
Witness is also missing the overhead, which is present in the C code.

roconnor-blockstream:
@uncomputable also noted on IRC that the disconnect combinator formula should be:

    overhead + 2 * bitSize(2^256 * A) + bitSize(B * C) + bitSize(B) + cost(s) + cost(t)

And the current implementation, both C and Haskell, is missing the `bitSize(B)` term.

roconnor-blockstream:
https://github.com/BlockstreamResearch/simplicity/blob/e5f74ca15ed553b769f3862d68b72271d1a98bd0/Haskell/Core/Simplicity/BitMachine/StaticAnalysis/Cost.hs#L43

In fact, this implementation in Haskell isn't even close to being correct.  The use of `reifyProxy` is pulling in completely incorrect types.  Unfortunately the type manipulations being done here are fairly opaque and a bit hard to make clear.

PR #166, opened 2023-07-06
Weight

roconnor-blockstream:
This PR does a couple of related things:

1. Makes the static analysis of memory and CPU bounds available for FFI bindings.
2. Generates expected CPU weight bounds for the generated test vectors.
3. Fixes the witness weight computation in Haskell, which addresses half of #159.

### DIFF
--- a/C/eval.c
+++ b/C/eval.c
@@ -814,8 +814,8 @@ static bool computeEvalTCOBound(memBound *dag_bound, const dag_node* dag, const 
   return true;
 }
 
-/* Run the Bit Machine on the well-typed Simplicity expression 'dag[len]'.
- * If 'NULL != input', initialize the active read frame's data with 'input[ROUND_UWORD(inputSize)]'.
+/* Run the Bit Machine on the well-typed Simplicity expression 'dag[len]' of type A |- B.
+ * If bitSize(A) > 0, initialize the active read frame's data with 'input[ROUND_UWORD(bitSize(A))]'.
  *
  * If malloc fails, returns 'SIMPLICITY_ERR_MALLOC'.
  * If static analysis results determines the bound on cpu requirements exceed the allowed budget, returns 'SIMPLICITY_ERR_EXEC_BUDGET'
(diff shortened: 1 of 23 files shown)


roconnor-blockstream:
@apoelstra only the C code needs to be reviewed.

I've shuffled around the commits quite a bit during development, so I do recommend giving it a pass through CI before looking at it too deeply.

Haskell/Core/Simplicity/BitMachine/StaticAnalysis/Cost.hs
```diff
@@ -40,7 +40,7 @@ instance Assert TermWeight where
 instance Witness TermWeight where
   witness _ = result
    where
-    result = TermWeight $ fromInteger (bitSizeR (reifyProxy result))
+    result = TermWeight $ overhead + milli (fromInteger . bitSizeR $ reifyProxy result)
```
uncomputable, line 43: 9dcdc08d5d8b073b4eef729c4d239dd630e566d8: This could be `TermWeight $ overhead + milli (bitSizeR (reifyProxy result))` right? That would make it the same as the for `iden` and for `comp`.

apoelstra:
utACK 56f15be54edd8692d1ceba21f8261890b68134dd. I have a mild preference for the "fix Cost.hs" commit to come first, before any computed values are introduced, avoiding committing then reverting bad numbers. But this is fine.

Will run CI when my box is free.

roconnor-blockstream:
Yeah regarding your "fix Cost.hs" preference.  Your choices are between being able to check that the regression test works, by swapping to adjacent commits, or doing your proposal and having to swap far apart commits, regenerate the failing data and dealing with the merge conflict that would arise.

I'm okay with either approach if you have a preference.  I don't see a way of getting the best of both solutions, other than maybe bizarre branches and merges.

Haskell/Core/Simplicity/BitMachine/StaticAnalysis/Cost.hs
roconnor-blockstream: Fixed.

apoelstra:
@roconnor-blockstream ah, yeah, good point. You do actually need to include the bad data for the test to still compile when it's swapped.

Ok, we'll keep things as is.

apoelstra (approved):
ACK 793c7b1fbc3146269732f1cffe715795ea0e9540

C/eval.c
```diff
@@ -809,56 +812,57 @@ static bool computeEvalTCOBound(memBound *dag_bound, const dag_node* dag, const
...
+                             );
+    *UWORDBound = (ubounded)ROUND_UWORD(inputSize) + (ubounded)ROUND_UWORD(outputSize)
+                + max(bound[len-1].extraUWORDBound[0], bound[len-1].extraUWORDBound[1]);
+    *frameBound = bound[len-1].extraFrameBound[0] + 2; /* add the initial input and output frames to the count. */
+    *costBound = bound[len-1].cost;
+  }
   free(bound);
-  return true;
+  return (maxCells <= *cellsBound) ? SIMPLICITY_ERR_EXEC_MEMORY
+       : (maxCost <= *costBound) ? SIMPLICITY_ERR_EXEC_BUDGET
```
roconnor-blockstream, line 828: I don't know how I missed it, but these inequalities were strict before and now they are non-strict, which is incorrect. I'll prepare a PR to correct this.

The written specification of "the bound on memory allocation requirements exceed the allowed limits" is just too vague.

PR #169, opened 2023-07-20
GenRustJets: rename `bitcoin_hashes` to `hashes`

apoelstra:
We made this change in rust-simplicity but forgot to update the generated code.


uncomputable (approved):
eb3c9cb8f0683d0ad8f37f38e4a8f99a8ea6bf5c ACK

PR #171, opened 2023-07-26
Include jet weights in their *MR calculations

roconnor-blockstream:
This means we are now commiting to their costs.


roconnor-blockstream:
Fixes #160 .

apoelstra (approved):
ACK ebd91218f7debdf9d8f7b51aeed9bfd242cfc483

PR #174, opened 2023-08-03
Iden8mebi

roconnor-blockstream:
Increase overhead for combinators.


apoelstra:
Will ACK shortly.

To check my understanding, this code always checks whether the actual cost bound is <= `maxCost`. But if `budget` is not provided, `maxCost` is set to `BOUNDED_MAX`, which is equal to the largest value that can be represented by a `ubounded` (which is `uint32_t`) and therefore this check always passes, and is effectively disabled.

apoelstra:
This feels a little bit too clever, as though you're using a sentinel value but then hiding the sentinel check inside implicit behavior of C integer types. I'm okay with it if you are but I think it could be implemented more clearly.

apoelstra (approved):
ACK d654859d9d3f366205d9bce5e06ef48c5edddbd8

roconnor-blockstream:
The specification indeed doesn't call out what happens when you pass in `BOUNDED_MAX`.  Would you like me to make it explicit.  I had left it out because it is, and should, only be invoked by testing and not by the main code.  I suppose I could make some sort of "for testing purposes only" pre and post condition claims.

apoelstra:
After discussion on IRC, I had two points of confusion:

* I had forgotten that `analyseBounds` clips all additions at `BOUNDED_MAX` (by its use of `bounded_add` rather than `+`)
* I had it in my head that `BOUNDED_MAX` was a somewhat-arbitrary parameter rather than "the maximum value of `ubounded`", because it scans similarly to `BUDGET_MAX` which *is* a somewhat-arbitrary parameter which is introduced in this PR as well.

In fact, the nature of `BOUNDED_MAX` isn't very important. What's important is that when you set `maxCost` to the `BOUNDED_MAX` constant used inside of `bounded_add`, it effectively disables the `maxCost < *costBound` check. I think we should add a comment near that check (it doesn't need to be part of the specification, it can just be a comment inside the function). The comment should simply say that because of the use of `bounded_add`, it is impossible for `*costBound` to exceed `BOUNDED_MAX`, so you can set `maxCost` to this value as a "disable bounds checking" sentinel.

apoelstra (approved):
ACK 56ac4027b9a2644ad6340b519e5a697cd22425f2

PR #177, opened 2023-08-10
Fee map

roconnor-blockstream:
Add a jet for lookup up transaction fees by assetid.


C/primitive/elements/env.c
```diff
@@ -515,7 +524,23 @@ extern transaction* elements_simplicity_mallocTransaction(const rawTransaction*
...
+     * Because the perm array entries are the same size as the feeOutputs array entries, it is safe to initialize one by one.
+     *
+     * In practical C implementations, the feeOutputs array entires are initalized to the same value as the perm array entries.
+     * In practical C implementations, this is a no-op, and generally compiliers are able to see this fact and eliminate this loop.
+     *
+     * We keep the loop in the code just to be pedantic.
+     */
+    for (size_t i = 0; i < numFees; ++i) {
+      static_assert(0 == offsetof(sigOutput, asset.data), "asset ID is not first field of sigOutput.");
+      feeOutputs[i] = (sigOutput*)(uintptr_t)(perm[i]);
```
real-or-random, line 561: Why cast to `(uintptr_t)` first?

C/primitive/elements/env.c
```diff
@@ -500,12 +501,20 @@ extern transaction* elements_simplicity_mallocTransaction(const rawTransaction*
     sha256_context ctx_outputSurjectionProofsHash = sha256_init(tx->outputSurjectionProofsHash.s);
     sha256_context ctx_outputsHash = sha256_init(tx->outputsHash.s);
     uint_fast32_t ix_fee = 0;
+
+    /* perm is a temporary array the same length (numFees) and size as feeOutputs.
+     * perm is used to initalize feeOutputs and is not used afterward.
+     * This makes it safe for perm to use the same memory allocation as feeOutputs.
+     */
+    static_assert(sizeof(const sha256_midstate*) == sizeof(sigOutput*), "Pointers (to sturctures) ought to have the same size.");
+    static_assert(alignof(const sha256_midstate*) == alignof(sigOutput*), "Pointers (to sturctures) ought to have the same alignment.");
```
real-or-random, line 530: typo: sturctures (2x)

C/primitive/elements/env.c
roconnor-blockstream: This suppresses warning about both casting away `const` and the warning about the change in pointer alignment.

I'll insert a comment about this.

C/primitive/elements/env.c
```diff
@@ -515,7 +524,23 @@ extern transaction* elements_simplicity_mallocTransaction(const rawTransaction*
...
+    simplicity_assert(ix_fee == numFees);
+    rsort(perm, numFees);
+
+    /* Initialize the feeOuputs array from the perm array.
+     * Because the perm array entries are the same size as the feeOutputs array entries, it is safe to initialize one by one.
+     *
+     * In practical C implementations, the feeOutputs array entires are initalized to the same value as the perm array entries.
+     * In practical C implementations, this is a no-op, and generally compiliers are able to see this fact and eliminate this loop.
+     *
+     * We keep the loop in the code just to be pedantic.
```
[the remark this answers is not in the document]
roconnor-blockstream: Yes.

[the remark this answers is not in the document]
roconnor-blockstream: I tried a couple of other avenues such as `rsort` taking a `void *` array or a `unsigned char *` array, but these changes didn't seem to make anything easier.  AFAICT, the C standard basically doesn't let you safely cast pointers to pointers into pointers to pointers of different types.  And indeed things get even less safer when if some of these types are non-structs.

roconnor-blockstream:
@apoelstra I believe this PR is ready now.  You may want to run it through your CI first to see if it passes all your checks.

C/primitive/elements/primitive.h
```diff
@@ -196,6 +197,7 @@ typedef struct sigInput {
 typedef struct transaction {
   const sigInput* input;
   const sigOutput* output;
+  const sigOutput* const * feeOutputs;
```
uncomputable, line 200: The `transaction` type is opaque so FFI is not affected, right? We build `transaction` from `rawTransaction` and then pass the opaque struct to the Simplicity runtime, without caring about its contents.

roconnor-blockstream: Yes. That is the intended API.

roconnor-blockstream: The intention is that the transaction value can be reused between validating multiple Simplicity inputs within one transaction.

apoelstra (approved):
ACK 473f3d42fc14bc76c03915d0d44de4a5403ebacb Carefully read all C code, though despite my care most of the comments explaining clever stuff were about things I would have missed entirely if there were no comments

PR #178, opened 2023-08-15
Output Is Fee

roconnor-blockstream:
Insert output-is-fee jet into element's jet catalogue.


apoelstra (approved):
ACK eadc29700a4d1c8e2a81d482ebc0f6f7dfd07a0b

PR #181, opened 2023-08-23
Update Benchmarks

roconnor-blockstream:
This PR also adds "overhead" on top of the jet cost.  This overhead isn't included in the CMR commitment, but is added on during the total cost calculation.

rust-simplicity needs to be updated with this change in overhead computation.


apoelstra (approved):
ACK fb209e3ee43e8e46834decad5c2f33511d5ea028

apoelstra:
I didn't test this against the Rust code -- in fact we know that it will break a bunch of our unit tests because the fixed merkle root targets will have changed.

sanket1729 (approved):
ACK fb209e3ee43e8e46834decad5c2f33511d5ea028

PR #183, opened 2023-08-24
Revamp radix sort

roconnor-blockstream:
The new implementation is non-recursive, ~~uses stack instead of heap (approximately 32Kibi),~~ and exposes an rsort function that doesn't abort early like was was used for `hasDuplicates`.

The new rsort functionality will be used to implement the "total-fees" jet which computes the total fees of any given asset id.

### DIFF
--- a/C/rsort.c
+++ b/C/rsort.c
@@ -109,15 +109,17 @@ static void sort_buckets(const sha256_midstate** a, uint32_t* restrict bucketSiz
 }
 
 /* Attempts to (partially) sort an array of pointers to 'sha256_midstate's in place in memcmp order.
- * If duplicate entries are found, the sorting is aborted and one of pointers to a duplicate entry is returned.
- * Otherwise if no duplicate entries are found 'NULL' is returned.
+ * If NULL == hasDuplicates then sorting is always run to completion.
+ * Otherwise, if NULL != hasDuplicates and if duplicate entries are found, the sorting is aborted and
+ * *hasDuplicates is set to one of pointers of a duplicate entry.
+ * If NULL != hasDuplicates and no duplicate entries are found, then *hasDuplicates is set to NULL.
  *
  * The time complexity of rsort is O('len').
  *
  * Precondition: For all 0 <= i < len, NULL != a[i];
  *               uint32_t stack[(CHAR_COUNT - 1)*(sizeof(*a)->s) + 1];
  */
-static const sha256_midstate* rsort(const sha256_midstate** a, uint_fast32_t len, uint32_t* stack) {
+static void rsort_ex(const sha256_midstate** a, uint_fast32_t len, const sha256_midstate** hasDuplicates, uint32_t* stack) {
   unsigned int depth = 0;
   size_t bucketCount[sizeof((*a)->s) + 1];
   size_t totalBucketCount = 1;
@@ -206,24 +208,48 @@ static const sha256_midstate* rsort(const sha256_midstate** a, uint_fast32_t len
         sort_buckets(a, bucketSize, bucketEdge, depth);
 
         depth++; bucketCount[depth] = CHAR_COUNT; totalBucketCount += CHAR_COUNT;
-      } else {
-        /* Early return when we are searching for duplicates and have found a bucket of size 2 or more
+        continue;
+      }
+      if (hasDuplicates) {
+        /* Early return if we are searching for duplicates and have found a bucket of size 2 or more
            whose "prefix" agree up to then entire length of the hash value, and hence are all identical.
          */
-        return a[begin];
+        *hasDuplicates = a[begin];
+        return;
       }
-    } else {
-      /* len - stack[totalBucketCount] < 2 */
-
-      /* When the last bucket size is 0 or 1 there is no sorting to do within the bucket.
-         It is already sorted, and since it is at the end we can decrease len.
-       */
-      len = stack[totalBucketCount];
     }
+    /* len - stack[totalBucketCount] < 2 || sizeof((*a)->s) == depth */
+
+    /* When the last bucket size is 0 or 1, or when the depth exceeds sizeof((*a)->s), there is no sorting to do within the bucket.
+       It is already sorted, and since it is at the end we can decrease len.
+    */
+    len = stack[totalBucketCount];
   }
   simplicity_assert(0 == len);
 
-  return NULL;
+  if (hasDuplicates) *hasDuplicates = NULL;
+}
+
+/* Sorts an array of pointers to 'sha256_midstate's in place in memcmp order.
+ * If malloc fails, returns false.
+ * Otherwise, returns true.
+ *
+ * The time complexity of rsort is O('len').
+ *
+ * We are sorting in memcmp order, which is the lexicographical order of the object representation, i.e. the order that one
+ * gets when casting 'sha256_midstate' to a 'unsigned char[]'. This representation is implementation defined, and will differ
+ * on big endian and little endian architectures.
+ *
+ * It is critical that the details of this order remain unobservable from the consensus rules.
+ *
+ * Precondition: For all 0 <= i < len, NULL != a[i];
+ */
+bool rsort(const sha256_midstate** a, uint_fast32_t len) {
+  uint32_t *stack = malloc(((CHAR_COUNT - 1)*(sizeof((*a)->s)) + 1) * sizeof(uint32_t));
+  if (!stack) return false;
+  rsort_ex(a, len, NULL, stack);
+  free(stack);
+  return true;
 }
 
 /* Searches for duplicates in an array of 'sha256_midstate's.
@@ -249,7 +275,9 @@ int hasDuplicates(const sha256_midstate* a, uint_fast32_t len) {
       perm[i] = a + i;
     }
 
-    result = NULL != rsort(perm, len, stack);
+    const sha256_midstate *duplicate;
+    rsort_ex(perm, len, &duplicate, stack);
+    result = NULL != duplicate;
   }
 
   free(perm);
--- a/C/rsort.c
+++ b/C/rsort.c
@@ -2,12 +2,18 @@
 
 #include <string.h>
 
-/* Return the 'i'th char of the internal representation of the midstate pointed to by a.
+/* Return the 'i'th char of the object representation of the midstate pointed to by a.
+ *
+ * In C, values are represented as 'unsigned char [sizeof(v)]' array.  However the exact
+ * specification of how this represenation works is implementation defined.
+ *
+ * For the 'uint32_t' values of 'sha256_midstate', the object representation of these values will differ
+ * between big endian and little endian archtectures.
  *
  * Precondition: NULL != a
  *               i < sizeof(a->s);
  */
-static unsigned char readLevel(const sha256_midstate* a, size_t i) {
+static unsigned char readIndex(const sha256_midstate* a, size_t i) {
   return ((const unsigned char*)(a->s))[i];
 }
 
@@ -27,11 +33,11 @@ static bool freq(size_t* result, const sha256_midstate * const * a, size_t len, 
   if (0 == len) return true;
 
   for (size_t i = 0; i < len - 1; ++i) {
-    result[readLevel(a[i],j)]++;
+    result[readIndex(a[i],j)]++;
   }
 
   /* Check the final iteration to see if the frequency is equal to 'len'. */
-  return len == ++result[readLevel(a[len-1],j)];
+  return len == ++result[readIndex(a[len-1],j)];
 }
 
 /* Given an array of bucket sizes, return an array of their cumulative sizes.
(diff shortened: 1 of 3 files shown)


apoelstra:
cc @real-or-random I am probably not going to be able to review this since I went on leave a bit earlier than expected. Could you take it on?

roconnor-blockstream:
Let's just leave this pending for the time being.  I'll either have Tim review it later, or you will be back.

real-or-random:
> Let's just leave this pending for the time being. I'll either have Tim review it later, or you will be back.

Okay, so I can't judge the priorities, of course. I'm happy to review this, but if you feel it can/should wait, just ping me again later.

C/rsort.c
```diff
@@ -70,64 +64,202 @@ static void swap(const sha256_midstate** a, const sha256_midstate** b) {
...
- * Precondition: size_t scratch[(level + 1)*CHAR_COUNT];
- *               For all 0 <= i < len, NULL != a[i];
- *               level <= sizeof((*a)->s);
+ * Precondition: For all 0 <= i < len, NULL != a[i];
  */
-const sha256_midstate* rsort(size_t* scratch, const sha256_midstate** a, size_t len, size_t level) {
-  /* An array of length 0 or 1 is sorted and without duplicates. */
-  if (len < 2) return NULL;
+static void rsort_ex(const sha256_midstate** a, uint_fast32_t len, const sha256_midstate** hasDuplicates) {
+  uint32_t stack[(CHAR_COUNT - 1)*(sizeof(*a)->s) + 1]; /* approximately 32 KiB */
```
roconnor-blockstream, line 115: There is an open question as to whether this should be allocated on the stack or on the heap.  Previously it was heap allocated.  In this PR it is stack allocated.

I ran this code through https://github.com/PeterMcKinnis/WorstCaseStack to get some perspective on this issue.
Note: I had to manually comment out jet evaluation.  Because jets are "dynamically dispatched" the stack analysis cannot know how much stack space jets will use.  So the numbers we have below are without any jets being used.  Also note: The analysis also doesn't know how much stack space `free`, `malloc` etc use, and the values also don't include whatever stack space is used by these functions.

The analysis of this PR suggests that `rsort` uses 34056 bytes of stack and `elements_simplicity_execSimplicity` uses 34784 bytes of stack.

If we were to heap allocate this, instead then `rsort` would use roughly  1432 bytes of stack and `elements_simplicity_execSimplicity` would use 2160 bytes of stack.

Relatedly, the worst case stack use of any of our current jets appears to be `check_sig_verify` which uses 5704 bytes of stack, with `secp256k1_ecmult` taking up 4632 bytes of stack

One thing to be mindful of is low-power hardware devices wanting to evaluate Simplicity.  For example Jade hardware had by default a very small stack, and they perform various stack tricks at the moment to support stack intensive operations required to do secp-zkp operations like verifying rangeproofs.

roconnor-blockstream: @apoelstra Unless you object, I think I'll update this PR this to being heap allocated.

apoelstra: Yeah, I think that's best. Staying under 8K stack will make a big difference to the portability of the code. And I don't think this library can work without malloc anyway, so there's no portability cost to heap allocation...just a code review/maintenance cost.

roconnor-blockstream: It is now heap allocated.

apoelstra:
In e7333674945a3b6e0b076f3f9de0e63dfc1cd678:

I'd prefer the comment on `rsort_ex` explicitly say that if `hasDuplicates` is NULL, then the sorting will not be aborted early. (I guess this is the obvious read for somebody who didn't know the old behavior, but I'd still prefer it be explicit.)

Done reviewing all but the long commit.

apoelstra:
The comment e733367 also makes a reference to "memcmp order".

apoelstra:
In 80a975b2d045769acf5b65552b055b63697a1a3d:

The commit message says that this eliminates malloc in favor of a large stack allocation, but this is out of date. Now we malloc a variable called `stack`.

roconnor-blockstream:
> The commit message says this is changing `rsort` to use "memcmp order" which is "implementation defined", but in fact you are changing the order to lexicographic, from right-to-left lexicographic. So you are moving from one specific ordering to a different one. The commit message should be changed.

Let's find a good way of phrasing this to make everyone happy (or unhappy).

`memcmp` order is the lexicographic order of the array when it is cast to an `unsigned char` array array.  However, how C casts a `uint32_t` to a `unsigned char` array is implementation defined, and in practice depends on the endianness of the architecture.  If the sort order were observable, this would cause a consensus failure when the code is compiled on different architectures.  However in our two applications, looking for duplicates, and performing a binary lookup, the underlying sort order is unobservable, so it is safe.

The previous order was also implementation defined in the same way, because again it was casting to an `unsigned char` array, which as noted above is implementation defined.

apoelstra:
Ok, other than the above nits (about comment and commit messages), this looks good to me. The algorithm is pretty complex, but with ChatGPT4's help I was able to step through the whole thing and understand it. Some very clever stuff.

I thought briefly about your stack bound and it seemed to be correct (or at least sufficient). Interestingly ChatGPT thinks that it could be tightened:

[The remainder of this post is a copy/paste from a chatgpt conversation. I apologize for the mangled formatting.]

You
Can you confirm that the precondition for `stack`, that it point to an allocation of size `(CHAR_COUNT - 1)*(sizeof(*a)->s) + 1`, is correct?
ChatGPT
ChatGPT
To confirm whether the precondition for stack in rsort_ex is correct, let's analyze its usage and the requirement of its size.

The stack in rsort_ex is used to keep track of the edges (or boundaries) of the buckets. This stack evolves as the radix sort algorithm progresses through each character (or byte) in the data structure. The size of stack needs to be sufficient to hold all potential bucket boundaries.

Analyzing the Stack Size Requirement
Maximum Depth: The maximum depth of sorting is sizeof((*a)->s). This is the size of the structure (or the key) being sorted. Each level of depth corresponds to one character (or byte) in the data structure.

Buckets at Each Depth: At each depth, there can be up to CHAR_COUNT buckets (presumably CHAR_COUNT is the number of distinct values a character can have, often 256 for byte-sized characters).

Stack Size Calculation:

For each depth, we need space for CHAR_COUNT bucket boundaries. However, the total number of bucket boundaries is less than CHAR_COUNT * sizeof((*a)->s) because the first depth doesn't split the array into buckets (it's the whole array). So, at most, we have CHAR_COUNT - 1 boundaries for each subsequent depth.
The total stack size then becomes (CHAR_COUNT - 1) * (sizeof((*a)->s) - 1). However, we also need one additional slot for the initial whole array (the 0th boundary).
Precondition Formula: (CHAR_COUNT - 1)*(sizeof(*a)->s) + 1

This formula seems to overestimate the required size. It considers CHAR_COUNT - 1 boundaries for each character in the data structure, including the first one, and then adds one more.
The correct formula should probably be (CHAR_COUNT - 1) * (sizeof((*a)->s) - 1) + 1. This formula accounts for the reduced number of boundaries at the first depth and adds one for the initial whole array boundary.
Conclusion
The provided precondition for stack seems to slightly overestimate the required size. It's not incorrect in the sense that it certainly allocates enough memory, but it allocates more than is strictly necessary. For most practical purposes, especially when CHAR_COUNT is 256 (as in byte-based radix sorts) and sizeof((*a)->s) is not extremely large, this overestimation might not be significant in terms of memory usage. However, for the sake of precision and optimal memory usage, revising the formula to `(CHAR_COUNT - 1) * (sizeof((*a)->s) - 1) + 1` would be more accurate. This revision properly accounts for the fact that the initial array doesn't need to be divided into buckets and thus reduces the stack size accordingly, while still ensuring there's enough space for all possible bucket boundaries at each level of depth.

It's important to note that such adjustments are essential in environments where memory efficiency is critical. In most desktop or server environments, the difference might not be impactful, but in memory-constrained environments like embedded systems, every byte counts. The provided formula errs on the side of caution by over-allocating slightly, which is a common practice to avoid buffer overruns, but understanding and optimizing such allocations can be beneficial in optimizing the overall memory footprint of the program.

roconnor-blockstream:
I didn't really read the ChatGPT nonsense very carefully, but what I can say is that performing the following patch

```
@@ -254,7 +254,7 @@ int hasDuplicates(const sha256_midstate* a, uint_fast32_t len) {
   static_assert(DAG_LEN_MAX <= SIZE_MAX / sizeof(const sha256_midstate*), "perm array too large.");
   simplicity_assert(len <= SIZE_MAX / sizeof(const sha256_midstate*));
   const sha256_midstate **perm = malloc(len * sizeof(const sha256_midstate*));
-  uint32_t *stack = malloc(((CHAR_COUNT - 1)*(sizeof((*perm)->s)) + 1) * sizeof(uint32_t));
+  uint32_t *stack = malloc(((CHAR_COUNT - 1)*(sizeof((*perm)->s) - 1) + 1) * sizeof(uint32_t));
   int result = perm && stack ? 0 : -1;
 
   if (0 <= result) {
```

causes `valgrind ./test` to fail with

```
valgrind: m_mallocfree.c:304 (get_bszB_as_is): Assertion 'bszB_lo == bszB_hi' failed.
valgrind: Heap block lo/hi size mismatch: lo = 31696, hi = 137438953504.
This is probably caused by your program erroneously writing past the
end of a heap block and corrupting heap metadata.  If you fix any
invalid writes reported by Memcheck, this assertion failure will
probably go away.  Please try that before reporting this as a bug.
```

I guess I still have a job.

In general, I fear ChatGPT might be more persuasive than correct when it comes to explaining its reasoning.

C/rsort.c
```diff
@@ -70,64 +64,209 @@ static void swap(const sha256_midstate** a, const sha256_midstate** b) {
...
+ *
+ * The time complexity of rsort is O('len').
+ *
+ * Precondition: For all 0 <= i < len, NULL != a[i];
+ */
+bool rsort(const sha256_midstate** a, uint_fast32_t len) {
+  uint32_t *stack = malloc(((CHAR_COUNT - 1)*(sizeof((*a)->s)) + 1) * sizeof(uint32_t));
+  if (!stack) return false;
+  rsort_ex(a, len, NULL, stack);
+  return true;
```
roconnor-blockstream, line 238: missing call to `free`.

roconnor-blockstream:
Similarly we get valgrind errors with

```
@@ -254,7 +254,7 @@ int hasDuplicates(const sha256_midstate* a, uint_fast32_t len) {
   static_assert(DAG_LEN_MAX <= SIZE_MAX / sizeof(const sha256_midstate*), "perm array too large.");
   simplicity_assert(len <= SIZE_MAX / sizeof(const sha256_midstate*));
   const sha256_midstate **perm = malloc(len * sizeof(const sha256_midstate*));
-  uint32_t *stack = malloc(((CHAR_COUNT - 1)*(sizeof((*perm)->s)) + 1) * sizeof(uint32_t));
+  uint32_t *stack = malloc(((CHAR_COUNT - 1)*(sizeof((*perm)->s))) * sizeof(uint32_t));
   int result = perm && stack ? 0 : -1;
 
   if (0 <= result) {
```

apoelstra:
Heh, good idea just trying the suggestion rather than trying to parse out its reasoning. In general:

* It was able to explain the bucketsort pretty effectively and I was able to understand the algorithm after reading its explanation.
* It struggled with `rsort_ex` but was able to describe the meaning of the variables well enough that I could figure out what was going on. But mostly its description of the algorithm was garbled.
* It suggested that there was a malloc without free, but worded it in a way that made me ignore it rather than noticing that a 5-line function was allocating memory then leaking it :/.
* It was very chatty and wordy and liked to comment about how unusual and specialized it was for a radix sort to be detecting duplicates, etc.

So overall quite helpful ... but yes, we both have jobs for the foreseeable future :P.

apoelstra:
And yes, the memory thing is quite persuasive and argument-shaped for something that is false.

roconnor-blockstream:
I believe I've addressed everything except possibly your concerns about the description of the sorting order.

apoelstra:
Ok, perfect. ACK everything except the commitment in 98b7a3ed0 which says

```
memcmp order is still implementation defined, however this order is not
    observable through the "hasDuplicate" or binary search lookup interfaces, so it
    is still safe to use in consensus.
```

My confusion is that this sounds like memcmp has implementation-defined behavior, which I don't believe is true, and contradicts your use of the term "memcmp order" to refer to a specific ordering.

roconnor-blockstream:
`memcmp` is a specific implementation defined order, which is the lexicographical order of the `unsigned char` array you get from casting a structure to an array.

Somewhere in between when you take a `struct foo *` and pass it into the `void *` argument of `memcmp` and when `memcmp` takes that `void *` and casts it to `unsigned char *`, something implementation defined happens.

I'm happy to reword this.  But the important thing, and the reason I'm making this sort order change, is that in I future PR I can do a binary table lookup using `memcmp`, which I couldn't do before because `rsort` didn't sort in `memcmp` order.

apoelstra:
Ok, I understand you. So in the specific case of sorting byte arrays, `memcmp` has no freedom. But for other types, such as ones with padding bytes or whatever, it does have freedom and the implementation makes specific choices.

I would appreciate rewording this a bit. I understand the motivation. I am just getting hung up on the phrase "implementation defined" when applied to an ordering that you clearly depend on.

roconnor-blockstream:
I don't depend on the order, beyond the fact that in a future PR I will depend on it being in the same order that `memcmp` uses, whatever that order is.

apoelstra:
This feels like a super subtle thing. I'm hearing:

* The ordering `memcmp` uses is implementation defined.
* **However** the ordering is constrained such that specifically on bytestrings, it must be lexicographic
* This PR manually orders bytestrings in the way that memcmp does, and we will call this "memcmp ordering"
* But just to psych the reviewer out, we will mention that "memcmp ordering" is actually implementation defined so our manual implementation needs to carefully stay within its constraints.

I think it'd be simpler if you just dropped the mention of implementation-definedness, used the phrase "memcmp ordering for bytestrings", and if you wanted to, mention that the memcmp ordering on non-bytestrings is implementation defined and we can't make assumptions about it.

roconnor-blockstream:
> * This PR manually orders bytestrings in the way that memcmp does, and we will call this "memcmp ordering"

Correction, "This PR manually orders `struct sha256_midstate`s, which is effectively a `uint32_t[8]` array, in the way that `memcmp` does..."

If `sha256_midstate` were (effectively) an unsigned char array, then everything would be well-defined, but it isn't (because sha256's implementation operates on 32-bit words).

roconnor-blockstream:
cppreference.com has some useful (and presumably standard language here).  memcmp orders objects lexicographically by their object representation.  But, aside from unsigned char arrays, object representations are all implementation defined.  (also something about bit-fields.)

apoelstra:
Ok -- so because your code works on `sha256_midstate`s, which hare not `unsigned char[]`s, and does not call `memcmp`, then it is incorrect to say that it uses "memcmp ordering"?

roconnor-blockstream:
I've added a few more comments after talking with @apoelstra out-of-band about this issue.

Hope this helps.

apoelstra:
Thanks! This helps a lot.

apoelstra (approved):
ACK dfb5b0cb7d85fe0a4f29d449f16b7f3011766823

uncomputable (commented):
dfb5b0cb7d85fe0a4f29d449f16b7f3011766823 The `hasDuplicates` unit tests successfully run on my end.

PR #184, opened 2023-08-30
GenRustJets: ElementsEnv with argument

uncomputable:
Fixes the Elements environment which now takes a type argument.

### DIFF
--- a/Haskell-Generate/GenRustJets.hs
+++ b/Haskell-Generate/GenRustJets.hs
@@ -290,6 +290,7 @@ rustJetImpl mod = vsep $
     ]
    where
     env | Nothing <- moduleName mod = "()"
+        | Just "Elements" == moduleName mod = "ElementsEnv<std::sync::Arc<elements::Transaction>>"
         | Just name <- moduleName mod = name ++ "Env"
     cEnv | Just "Elements" == moduleName mod = "CElementsTxEnv"
          | otherwise = "()"


sanket1729:
@uncomputable, this does not work if I use the generated file. 

Cannot offer review on the Haskell part, but this is the error in rust-simplicity. 
```rust
   Compiling simplicity-sys v0.1.0 (/home/sanket1729/rust-simplicity/simplicity-sys)
   Compiling simplicity v0.1.0 (/home/sanket1729/rust-simplicity)
error[E0107]: missing generics for struct `ElementsEnv`
   --> src/jet/init/elements.rs:317:24
    |
317 |     type Environment = ElementsEnv;
    |                        ^^^^^^^^^^^ expected 1 generic argument
    |
note: struct defined here, with 1 generic parameter: `T`
   --> src/jet/elements/environment.rs:61:12
    |
61  | pub struct ElementsEnv<T: Deref<Target = elements::Transaction>> {
    |            ^^^^^^^^^^^ -
help: add missing generic argument
    |
317 |     type Environment = ElementsEnv<T>;
    |                                   +++

For more information about this error, try `rustc --explain E0107`.
error: could not compile `simplicity` (lib) due to previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `simplicity` (lib test) due to previous error
```

sanket1729 (approved):
sorry, my previous comment was incorrect. The new code builds correctly.

PR #186, opened 2023-09-13
Fix compatibility with stable nixpkgs

uncomputable:
`nix-shell` no longer works on nixpkgs 23.05 (latest stable) since 4c32bcf53c59eabc80c97f607754d784b67311db. This PR partially reverts the changes of this commit as a fix.


roconnor-blockstream:
Actually this seems to have broken `nix-shell` when using the unstable channel.

uncomputable:
Oops. You can revert the shell to unstable and I can use a separate shell for stable, if we cannot unify both worlds.

PR #187, opened 2023-09-20
Have nix-shell work on both stable and unstable nixpkgs

roconnor-blockstream:
Apparently I incorrectly tested #186 and in fact it breaks running `nix-shell` on the unstable nixpkgs channel.

I expect this change to work for both the stable and unstable channels.


uncomputable:
Running `nix-shell --arg coq false --arg c false` (what I usually do) gives me the same shell as the one on `master`. Nix uses the same cache. If this is the expected behavior, then it seems to work on stable.

Issue #190, opened 2023-10-18
Error in Boolean Jet decoding


roconnor-blockstream:
https://github.com/BlockstreamResearch/simplicity/pull/185 has an error in it where the boolean jets are decoded with code ;0`; however it should be code `1` instead.

I expect that eventually this will be found when fuzz testing (and updated) rust-simplicity.

Still, we should add a Haskell FFI binding to this function and test that every jet can be decoded correctly.  Then fix this error.

PR #191, opened 2023-10-18
Fix decoding error in Boolean jets

roconnor-blockstream:
The decoding of the boolean case should be at code 1, not code 0.

I've changed the Haskell code so that putPositive crashes when given a non-positive value, rather than silently putting 1 value.  This probably would have prevented the error, and indeed the Haskell tests fail with only this change to putPositive.

Further mitigations to be done in a later commit are suggested in #190.


roconnor-blockstream:
There is no way any review could have reasonably been expected to catch this error of mine in #185.  This PR adds some mitigations, and further mitigations should be added later per #190.

uncomputable:
utACK 04361d7. The changes are minimal and the code compiles on my machine. I didn't find any occurrences of the old encoding in the code.

We will have to update rust-simplicity for support.

uncomputable (approved):

roconnor-blockstream:
The rust encoding/decoding code is autogenerated and ought to be/will be correct.  In fact, the fuzz tests between the (correct) rust code and the incorrect version of the C code would likely fail if we were to run them.

real-or-random:
I don't think I can provide a meaningful review here, right now. I guess that's fine if you two agree that this change is correct.

Still, it would be great if you could tell me more about codes. What are we encoding and decoding here, and how is it done (roughly)?

roconnor-blockstream:
@real-or-random Simplicity expressions trees are squished into a DAG by identifying identical sub-expression.  The DAG is serialized from leaf to root as a sequence of nodes.  Nodes are encoded using a prefix-free, self-delimiting encoding.  There are different node types for different types of expressions, `case`, `pair`, each possible jet, etc.  Nodes link to their children via encoding the relative offset (DAGs can only point away from the root) from their node's index to their child's index via a strictly positive number (DAGs can only point away from the root), which is itself encoded using a fancy prefix-free, self-delimiting encoding.

There are many jets, and they are catalogued into a hierarchy.  Each jet is encoded by a prefix indicating this is a jet, followed by one or more strictly positive integers that give an index into this catalogue.

Anyhow, I incorrectly typed the wrong value for these jets for their catalogue value.  In fact, they were manifestly incorrect because their value wasn't even strictly positive.

On the Haskell side, I did recently add a test that checks that every jet can be decoded from their encoding.  Normally you'd expect such an error to be caught by this test, however, before this PR, if you asked the Haskell encoder for strictly positive numbers, to encode an non-positive number, it would just silently output the encoding for `1` instead.  Which in this case was in fact the correct encoding.

Then the C code was copied from the incorrect text of the Haskell encoding.  Due to inadequate testing of the C code (see #190) the C decoding isn't covered by the tests.  It is covered by the Rust fuzz tests, so would have likely eventually been uncovered.  But still, we need to have better testing of the C code here.  And actually I'd like to generated this C decoder code automatically since it is tedious to maintain by hand.

real-or-random (approved):
Okay, makes sense on a high level...

utACK 04361d7a3565d594f058cfbb34b14bfc46901a9e whatever this means with my limited knowledge here :)

roconnor-blockstream:
Your utACK is more about making sure that we have considered the corner cases of C semantics that an expert like you might be able to uncover that a non-expert might overlook.  I mean, we are just changing labels in a switch statement, so I'm not expecting you to find anything sneaky.  This is more of a policy that all C code changes require two reviewers, because no matter how trivial the C code changes might seem, C semantics is just waiting to mess something up.

real-or-random:
Okay, yes, thanks for reminding me about this. And yes, I'm rather confident that this PR does not introduce unexpected semantic changes. :)

PR #192, opened 2023-10-24
Refactor name generation

roconnor-blockstream:
At the same time we generalize it to support names with underscores, which we will be soon adding.

This PR doesn't change the current output of GenPrimitives or GenRustJets.


uncomputable (approved):
ACK 7a6ba113dff8d77e65a5f35d2ac613e219633125. I can confirm that the output for GenPrimitive and GenRustJets is the same as on master.

PR #193, opened 2023-10-26
Shift jets

roconnor-blockstream:
Add a first batch of jets related to bit shifting (by static amounts) and similar.


roconnor-blockstream:
As usual, strictly speaking, only the C code (under the C directory), needs review.

C/jets.c
```diff
@@ -405,6 +405,62 @@ RIGHTMOST_(64,8)
...
+LEFT_PAD_LOW_1_(8)
+LEFT_PAD_LOW_1_(16)
+LEFT_PAD_LOW_1_(32)
+LEFT_PAD_LOW_1_(64)
+
+#define LEFT_PAD_LOW_(bitsN, bitsM)                                                    \
+/* left_pad_low_n_m : TWO^n |- TWO^m */                                                \
+bool left_pad_low_##bitsN##_##bitsM(frameItem* dst, frameItem src, const txEnv* env) { \
+  (void) env; /* env is unused. */                                                     \
+  for(int i = 0; i < (bitsM)/(bitsN) - 1; ++i) { write##bitsN(dst, 0); }               \
```
uncomputable, line 425: a9f2c88d13783ee145c73d1545ee7a201edf01c9: `(bitsM)/(bitsN)`

This is a potential divide by zero. We have to use the macro correctly to exclude this case. Is this a problem? Should we add a warning in the docs?

C/jets.c
```diff
@@ -405,6 +405,62 @@ RIGHTMOST_(64,8)
...
+RIGHT_PAD_LOW_1_(16)
+RIGHT_PAD_LOW_1_(32)
+RIGHT_PAD_LOW_1_(64)
+
+#define RIGHT_PAD_LOW_(bitsN, bitsM)                                                    \
+/* right_pad_low_n_m : TWO^n |- TWO^m */                                                \
+bool right_pad_low_##bitsN##_##bitsM(frameItem* dst, frameItem src, const txEnv* env) { \
+  (void) env; /* env is unused. */                                                      \
+  copyBits(dst, &src, (bitsN));                                                         \
+  for(int i = 0; i < (bitsM)/(bitsN) - 1; ++i) { write##bitsN(dst, 0); }                \
```
uncomputable, line 512: a9f2c88d13783ee145c73d1545ee7a201edf01c9: Again `(bitsM)/(bitsN)`

C/jets.c
```diff
@@ -405,6 +405,62 @@ RIGHTMOST_(64,8)
...
+LEFT_PAD_LOW_(16,32)
+LEFT_PAD_LOW_(16,64)
+LEFT_PAD_LOW_(32,64)
+
+#define RIGHT_PAD_LOW_1_(bitsM)                                                          \
+/* right_pad_low_1_m : TWO |- TWO^m */                                                   \
+bool right_pad_low_1_##bitsM(frameItem* dst, frameItem src, const txEnv* env) {          \
+  (void) env; /* env is unused. */                                                       \
+  bool bit = readBit(&src);                                                              \
+  write##bitsM(dst, (uint_fast##bitsM##_t)((uint_fast##bitsM##_t)bit << ((bitsM) - 1))); \
```
uncomputable, line 499: a9f2c88d13783ee145c73d1545ee7a201edf01c9: `... << ((bitsM) - 1)`

If `bitsM` is zero, then we shift by negative amounts.

uncomputable:
The jets defined in 416c0c53c54fcc862f4a80db94fe23f5cd743ad9 and e18e84dacb61a867707c86b03164a03834c63932 have similar corner cases with undefined behavior.

I'm probably being pedantic. It is so easy to get undefined behavior in C.
```
copyBits(dst, &src, (bitsN) + (bitsM));
```
What if `(bitsN)+(bitsM)` exceeds the maximum value? And so on.

uncomputable (commented):
e18e84dacb61a867707c86b03164a03834c63932: The C code looks good with some potential corner cases (see comments). I have skimmed the Haskell code and run the testsuite. I can confirm that the .inc files were correctly generated.

C/jets.c
real-or-random: Yes, but again we'd call `write0`.

By the way, I don't think you'll need to cast the result of the `<<` operation. (But it's not wrong either.)

C/jets.c
```diff
@@ -433,6 +433,34 @@ LEFT_PAD_LOW_(16,32)
 LEFT_PAD_LOW_(16,64)
 LEFT_PAD_LOW_(32,64)
 
+#define LEFT_PAD_HIGH_1_(bitsM)                                                 \
+/* left_pad_high_1_m : TWO |- TWO^m */                                          \
+bool left_pad_high_1_##bitsM(frameItem* dst, frameItem src, const txEnv* env) { \
+  (void) env; /* env is unused. */                                              \
+  for(int i = 0; i < (bitsM) - 1; i++) { writeBit(dst, true); }                 \
```
real-or-random, line 440: style nit: sometimes you use `i++`, and sometimes `++i`

real-or-random (commented):
ACK mod nits.

I haven't really looked at the moved code.

roconnor-blockstream:
I've pushed some extra guard-rails for this PR.

real-or-random (approved):
ACK 9aba2e78401a0a52edfba54c9742a71471534b79 C stuff is sound

The static asserts are a good idea.

uncomputable (approved):
ACK 9aba2e78401a0a52edfba54c9742a71471534b79

PR #197, opened 2023-11-02
Rename bitstream errors

uncomputable:
The new name and error message clarify what the error does.

### DIFF
--- a/C/include/simplicity/errorCodes.h
+++ b/C/include/simplicity/errorCodes.h
@@ -20,7 +20,7 @@ typedef enum {
   SIMPLICITY_ERR_STOP_CODE = -10,
   SIMPLICITY_ERR_HIDDEN = -12,
   SIMPLICITY_ERR_BITSTREAM_UNUSED_BYTES = -14,
-  SIMPLICITY_ERR_BITSTREAM_UNUSED_BITS = -16,
+  SIMPLICITY_ERR_BITSTREAM_ILLEGAL_PADDING = -16,
   SIMPLICITY_ERR_TYPE_INFERENCE_UNIFICATION = -18,
   SIMPLICITY_ERR_TYPE_INFERENCE_OCCURS_CHECK = -20,
   SIMPLICITY_ERR_TYPE_INFERENCE_NOT_PROGRAM = -22,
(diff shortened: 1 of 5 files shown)


roconnor-blockstream (approved):
LGTM

uncomputable:
@real-or-random's comment made me think more about what the bitstream-related errors do. I pushed two more commits.

If my understanding of SIMPLICITY_ERR_WITNESS_UNUSED_BITS is correct, then the new name should be more descriptive than the old one. I don't have strong opinions about SIMPLICITY_ERR_BITSTREAM_UNUSED_BYTES.

C/include/simplicity/errorCodes.h
```diff
@@ -79,8 +79,8 @@ static inline const char * SIMPLICITY_ERR_MSG(simplicity_err err) {
     return "Expression not unit to unit";
   case SIMPLICITY_ERR_WITNESS_EOF:
     return "Unexpected end of witness block";
-  case SIMPLICITY_ERR_WITNESS_UNUSED_BITS:
-    return "Unused data at the end of the witness block";
+  case SIMPLICITY_ERR_WITNESS_LENGTH_MISMATCH:
+    return "Witness block is longer than declared";
```
roconnor-blockstream, line 83: It isn't longer than declared.  The prefix of the witness block declares how long it is and the decoded block is always that long.  However the witness block contains a sequence of witness values, and it could be that the witness block is longer than necessary, in which case this error is raised.  It could also be shorter than necessary, in which case a `SIMPLICITY_ERR_WITNESS_EOF` is raised.

uncomputable: I see. We always decode as many bits as declared at the start of the witness block. _(The maximum length is bounded by 2^31 exclusive to prevent DDoS.)_ If there are not enough bits, then we throw an EOF error. After parsing the witness, we check if there are trailing bits left, in which case we throw UNUSED_BITS.

uncomputable:
I revised by final commit. In the end, I rename `*_UNUSED_{BITS,BYTES}` to `*_TRAILING_{BITS,BYTES}`. I think this makes the errors slightly clearer, but I don't have strong feelings about them.

real-or-random (approved):
(trivial) ACK 8855fda75bf2c1bc52c5912180ce9fad93371ed2 no semantic subtleties here

roconnor-blockstream (approved):
LGTM

PR #198, opened 2023-11-06
Optimize Word Jet CMR calculation

### LINKED ISSUE #196: Verify the costs of computing Word Jet CMRs
The `word("2^23 zero bits") ; unit` regression test seems to take a few seconds to run.  We need to make sure that the assigned cost is compatible with the runtime of this computation.  (Possibly by improving the Word jet CMR computation itself.)

roconnor-blockstream:
Fixes #196.


C/test.c
```diff
@@ -528,6 +528,12 @@ static void regression_tests(void) {
     diff = (double)(end - start) / CLOCKS_PER_SEC;
     bound = (double)(sizeof_regression3) * secondsPerWU;
     printf("cpu_time_used by regression3: %f s.  (Should be less than %f s.)\n", diff, bound);
+    if (diff <= bound) {
+      successes++;
+    } else {
+      failures++;
+      printf("regression3 took too long.\n");
+    }
```
uncomputable, line 536: dd9a973a1a1eecdbe76573423c242d5ae0f15f1e: The outcome of this check depends on the machine it is run on. Is this OK for our CI? If regression3 takes too long, then this seems like a softer error than if, say, the cmr was falsely computed.

C/bitstring.h
```diff
@@ -32,4 +32,21 @@ static inline bool getBit(const bitstring *s, size_t n) {
...
+  simplicity_assert(n <= s->len - 8);
+  size_t total_offset = s->offset + n;
+  if (total_offset % CHAR_BIT <= CHAR_BIT - 8) {
+    return (uint_fast8_t)(0xff & (s->arr[total_offset / CHAR_BIT] >> (CHAR_BIT - 8 - (total_offset % CHAR_BIT))));
+  } else {
+    /* CHAR_BIT < total_offset % CHAR_BIT + 8 */
+    return (uint_fast8_t)(0xff & (((1U * s->arr[total_offset / CHAR_BIT]) << (total_offset % CHAR_BIT - (CHAR_BIT - 8)))
+                                  | (s->arr[total_offset / CHAR_BIT + 1] >> (CHAR_BIT - (total_offset % CHAR_BIT - (CHAR_BIT - 8))))));
+  }
+}
```
uncomputable, line 51: @real-or-random Can you take a look at this? I find this difficult to read, even though the method only reads a single byte from a bitstring.

uncomputable (approved):
ACK dd9a973a1a1eecdbe76573423c242d5ae0f15f1e Looks good to me. The tests run and I could reproduce the precomputed data.

roconnor-blockstream:
@uncomputable This is a draft PR.  I don't want it reviewed yet.  I only posted it because you asked to see it.

roconnor-blockstream:
@real-or-random This might be harder to review, so let be know if you have any questions.

C/test.c
roconnor-blockstream: I'm torn on the issue because making the test based on time means it is inherently a flaky test.  On the other hand, I do want to know if it ever fails and not just ignore the issue.

uncomputable (approved):
ACK 4f3efc917edc9bf9cdd9ec0dc6bd4b601a083604 Minor changes from the previous state. `getByte()` seems correct, but someone should take a closer look at it.

C/dag.c
```diff
@@ -97,31 +97,39 @@ static sha256_midstate mkJetCMR(uint32_t *imr, uint_fast64_t weight) {
...
+     case 1: i = 2 + ((1U * getBit(value, 0) << 1) | getBit(value, 1)); break;
+     case 2: i = 6 + ((1U * getBit(value, 0) << 3) | (1U * getBit(value, 1) << 2) | (1U * getBit(value, 2) << 1) | getBit(value, 3)); break;
+    }
+    memcpy(stack_ptr, &word_cmr[i], sizeof(uint32_t[8]));
+  } else {
+    for(size_t i = 0; i < value->len >> 3; ++i) {
+      /* stack_ptr == stack + 8*<count of the number of set bits in the value i> */
+      stack_ptr += 8;
+      memcpy(stack_ptr, &word_cmr[22 + getByte(value, 8*i)], sizeof(uint32_t[8]));
+      /* This inner for loop runs in ammortized constant time. */
```
real-or-random, line 126: nit:
```suggestion
      /* This inner for loop runs in amortized constant time. */
```

C/dag.c
```diff
@@ -97,31 +97,39 @@ static sha256_midstate mkJetCMR(uint32_t *imr, uint_fast64_t weight) {
...
-    memcpy(stack_ptr, &bit_cmr[getBit(value, i)], sizeof(uint32_t[8]));
-    /* This inner for loop runs in ammortized constant time. */
-    for (size_t j = i; j & 1; j = j >> 1) {
-      sha256_midstate pair = cmrIV(PAIR);
-      stack_ptr -= 8;
-      sha256_compression(pair.s, stack_ptr);
-      memcpy(stack_ptr, pair.s, sizeof(uint32_t[8]));
+    size_t i;
+    switch(n) {
+     case 0: i = getBit(value, 0); break;
```
real-or-random, line 116: nit: Indentation is just 1 column here. Is this on purpose?

C/dag.c
```diff
@@ -97,31 +97,39 @@ static sha256_midstate mkJetCMR(uint32_t *imr, uint_fast64_t weight) {
...
+     case 1: i = 2 + ((1U * getBit(value, 0) << 1) | getBit(value, 1)); break;
+     case 2: i = 6 + ((1U * getBit(value, 0) << 3) | (1U * getBit(value, 1) << 2) | (1U * getBit(value, 2) << 1) | getBit(value, 3)); break;
+    }
+    memcpy(stack_ptr, &word_cmr[i], sizeof(uint32_t[8]));
+  } else {
+    for(size_t i = 0; i < value->len >> 3; ++i) {
```
real-or-random, line 122: nit:
```suggestion
    for  (size_t i = 0; i < value->len >> 3; ++i) {
```
(at least you write `for (` further down)

C/test.c
real-or-random: Hm, yes, having a flaky test may be annoying for users and distributors, too. But I don't have a great suggestion to improve it. (You could count CPU cycles of just the test thread, but this is just less flaky and probably much less portable.)

real-or-random (approved):
ACK 4f3efc917edc9bf9cdd9ec0dc6bd4b601a083604

C/bitstring.h
real-or-random: It looks correct to me. (I think it would be easier to reason about it if we could assume `CHAR_BIT == 8`, but I assume there's a reason why you don't want to make this assumption.)

C/dag.c
roconnor-blockstream: I guess that is my style.

C/test.c
roconnor-blockstream: I've opened issue #202.

real-or-random (commented):
ACK 7162e245d92a550b43cfd3e443d28158aec1bd90

uncomputable (approved):
ACK 7162e245d92a550b43cfd3e443d28158aec1bd90

Issue #194, opened 2023-10-30
Idem8mebi trailing bytes


uncomputable:
The [program of the idem8mebi test](https://github.com/BlockstreamResearch/simplicity/blob/69eb1719b6a13a4346abde9b221eebc99002cc6f/C/test.c#L278) fails in Elements with the error `SIMPLICITY_BITSTREAM_UNUSED_BYTES`.

These are the hardcoded program bytes:
```
e108000000000000000000000000000000000000000000000000000000000000004000
```

I rewrote the program in Rust to get the following bytes:
```
e108000000000000000000000000000000000000000000
```

The program produced by Rust doesn't fail in Elements.

Here is the human encoding that Rust used as input:

```
id0 := iden
cp0 := comp id0 id0
cp1 := comp cp0 cp0
cp2 := comp cp1 cp1
cp3 := comp cp2 cp2
cp4 := comp cp3 cp3
cp5 := comp cp4 cp4
cp6 := comp cp5 cp5
cp7 := comp cp6 cp6
cp8 := comp cp7 cp7
cp9 := comp cp8 cp8
cp10 := comp cp9 cp9
cp11 := comp cp10 cp10
cp12 := comp cp11 cp11
cp13 := comp cp12 cp12
cp14 := comp cp13 cp13
cp15 := comp cp14 cp14
cp16 := comp cp15 cp15
cp17 := comp cp16 cp16
cp18 := comp cp17 cp17
cp19 := comp cp18 cp18
cp20 := comp cp19 cp19
cp21 := comp cp20 cp20
main := comp cp21 cp21
```

roconnor-blockstream:
This program came from @apoelstra 's fuzz tester.  I had thought we hand normalized it before putting it into a test vector, but it seems from the IRC logs circa 2023-07-26 that we only ran it through the decoder, rather than also reencoding it.  So that is how this error came to be.

roconnor-blockstream:
I'll see if it is appropriate to update the test harness to fail with trailing bytes, and then trim the test vector appropriately.

PR #199, opened 2023-11-07
Trim extra data from iden8mebi test case

roconnor-blockstream:
This test case was originally generated through a fuzzer. Apparently the resulting test vector wasn't normalized when added to test.c, and extraneous data was left tacked onto the end.


uncomputable (approved):
ACK 99b438446dba620dd37e0bb509bc4caaa438a334 The program bytes look good and the tests run.

roconnor-blockstream:
Fixed #194

PR #203, opened 2023-11-15
Shift jets 2

roconnor-blockstream:
Add jets for dynamic shifting and rotating.


C/frame.c
```diff
@@ -1,13 +1,13 @@
 #include "frame.h"
 
-#define READ_(bits)                                                                                                           \
+#define READ_(bits,size)                                                    \
```
uncomputable, line 3: 1f245eccc7593859bcdcbfe24d0c09f4c8443867: We should add an explanation to the docs that `size` is the bit width of the target data type which is returned. `bits ≤ size` must hold. We could add a static assert.

Simplicity-TR.tm
```diff
@@ -9078,29 +9078,52 @@
...
+  <subsubsection|<samp|left-shift-with>>
+
+  Left shift by a given amount. Bits are filled with the provided value.
+
+  <\math>
+    <rep|<text|<samp|'left-shift-with'>><rsub|2<rsup|n>>|>\<assign\><verbatim|<around*|[|110|]>><rsub|<2>>\<cdummy\><rep|<value|subsection-nr>|>\<cdummy\><rep|<value|subsubsection-nr>|>\<cdummy\><rep|n+1|>
+  </math>
+
+  <math|<text|<samp|left-shift-with>><rsub|2<rsup|n>>
+  :<2>\<times\><around*|(|<2><rsup|2<rsup|<around*|\<lceil\>|lg<around*|(|n|)>|\<rceil\>>>>\<times\><2><rsup|2<rsup|n>>|)>\<vdash\><2><rsup|2<rsup|n>>>
```
uncomputable, line 9090: 1f245eccc7593859bcdcbfe24d0c09f4c8443867: `2 × (2^{2^ceil(lg n)}) × 2^{2^n} → 2^{2^n}`
Does `lg n` mean `log10 n`? That is surprising to me. I expected something like `2 × (2^n × 2^{2^n} → 2^{2^n}`

C/jets.c
```diff
@@ -589,6 +589,72 @@ RIGHT_EXTEND_(16,32)
 RIGHT_EXTEND_(16,64)
 RIGHT_EXTEND_(32,64)
 
+#define LEFT_SHIFT_(log, bits)                                                           \
+static inline void left_shift_helper_##bits(bool with, frameItem* dst, frameItem *src) { \
+  uint_fast8_t amt = read##log(src);                                                     \
```
uncomputable, line 594: 1f245eccc7593859bcdcbfe24d0c09f4c8443867 There will be a compiler error if `log > 8`. This will not break any binaries, but it might be confusing.

C/jets.c
```diff
@@ -589,6 +589,72 @@ RIGHT_EXTEND_(16,32)
...
+  return true;                                                                           \
+}
+LEFT_SHIFT_(4,8)
+LEFT_SHIFT_(4,16)
+LEFT_SHIFT_(8,32)
+LEFT_SHIFT_(8,64)
+
+#define RIGHT_SHIFT_(log, bits)                                                           \
+static inline void right_shift_helper_##bits(bool with, frameItem* dst, frameItem *src) { \
+  uint_fast8_t amt = read##log(src);                                                      \
```
uncomputable, line 627: 1f245eccc7593859bcdcbfe24d0c09f4c8443867: Ditto

uncomputable (commented):
36b2df29d61abca39846f651cec428c25fb6d3f6 I regenerated the autogenerated files and ran the tests

C/jets.c
real-or-random: I fail to see why there is a compiler error.

But yeah, this code assumes that `log` <= 64 (because otherwise truncation will happen.) Perhaps that should be asserted.

C/jets.c
```diff
@@ -655,6 +655,38 @@ RIGHT_SHIFT_(4,16)
 RIGHT_SHIFT_(8,32)
 RIGHT_SHIFT_(8,64)
 
+#define ROTATE_(log, bits)                                                                     \
+/* Precondition: 0 <= amt < bits */                                                            \
```
real-or-random, line 659: nit: could be asserted at runtime

roconnor-blockstream: Normally I don't add asserts to jets.

We cannot both have policy of keeping asserts in production and putting asserts in code that is to be executed as quickly as possible.

That said, I see I do use `simplicity_debug_assert` at one point, which is an assert that is removed when `PRODUCTION`, so I suppose I could use that if you think I should.

Simplicity-TR.tm
roconnor-blockstream: `lg` is the standard CS notation for base 2 logarithms.

roconnor-blockstream: Apparently the ISO wants us to use `lb` which is a notation I have never seen before: https://en.wikipedia.org/wiki/Binary_logarithm#Notation

C/jets.c
roconnor-blockstream: I think I would prefer not to do this.  If anything I think I would prefer to remove the existing `simplicity_debug_assert` that is used elsewhere.

roconnor-blockstream:
@apoelstra you do not have to review this PR, but you may wish to run it through your CI.

real-or-random (approved):
utACK d946adfba9a2d9b42739932a92421732f4478514

real-or-random:
Ha, these warnings are surprisingly stupid. 

> ```
> jets.c:684:54: error: conversion from 'int' to 'uint_fast8_t' {aka 'unsigned char'} may change valu>
> ```

Except that this is the compile-time constant 8, which AFAIK representable in `uint_fast8_t`.

apoelstra:
Would you like me to dig into my scripts and figure out the exact invocation to get these warnings? They seemed pretty self-contained so I just copy/pasted them.

roconnor-blockstream:
Fixed the GCC warnings

apoelstra (approved):
ACK da9f7f51904be2f7577305bedc6c7e100503903c

roconnor-blockstream:
rebased

apoelstra (approved):
ACK 925b4a7b67edc1f38b425abdabaa361bceaf6ae5

roconnor-blockstream:
@real-or-random Do you want to finish off this review?  It should be the last of it.

real-or-random (approved):
utACK 925b4a7b67edc1f38b425abdabaa361bceaf6ae5

(It was on my list for next week, but ok, I had not expected that the diff to the previous version is so easy to review.)

PR #207, opened 2023-11-26
WASM support

### LINKED ISSUE #206: Not WASM compatible
The current C code cannot be compiled for WASM, which prevents Simplicity to be run in the browser. rust-simplicity depends on C FFI, so it has the same problem. For comparison, libsecp can be compiled.

We have to remove incompatible standard library headers. So far it looks like these headers cause problems:

- stdio.h
- assert.h

uncomputable:
Fixes #206


uncomputable:
@roconnor-blockstream I updated this PR to do minimal changes only. Other WASM fixes will come in follow-up PRs.

roconnor-blockstream:
This can now be rebased.  It will probably help with Andrew's CI.

apoelstra (approved):
ACK bd3a322ca7825515ced3a5f969d7d2912f6c675a

PR #210, opened 2023-12-20
Fix C headers

uncomputable:
I used a bash script to detect C headers with incomplete dependencies. We could add it to the repo if you want.

The script can be run inside `nix-shell --arg coq false --arg haskell false` from the project root.

```bash
compile_header() {
    # Relative header path without leading ./
    local header="${1#./}"
    
    echo "#include \"$header\"" > "tmp.c"

    if ! "$CC" -c "tmp.c" -I include -o /tmp/dummy_out; then
        exit 1
    fi
}

# Make `compile_header` available for `find`
export -f compile_header

cd C
# Compile each header and quit on the first failure
# Exclude secp256k1 directory
find . -path "./secp256k1" -prune -o -name "*.h" \( -exec bash -c 'compile_header "$0"' {} \; -o -quit \)
rm tmp.c
cd ..
```

### DIFF
--- a/C/bitstring.h
+++ b/C/bitstring.h
@@ -2,6 +2,8 @@
 #ifndef SIMPLICITY_BITSTRING_H
 #define SIMPLICITY_BITSTRING_H
 
+#include <stddef.h>
+#include <stdint.h>
 #include <limits.h>
 #include <stdbool.h>
 #include "simplicity_assert.h"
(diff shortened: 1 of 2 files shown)


apoelstra (approved):
ACK ab86e5a55bcc80fd7af04bd522c5ed1df902bdeb

roconnor-blockstream (changes requested):
LGTM.  Needs to be rebased.

uncomputable:
Rebased

apoelstra (approved):
ACK f319952a8f2ae205889c713fd371b5fed38c7b9b

roconnor-blockstream (approved):
LGTM

PR #214, opened 2024-01-05
Profiler support

roconnor-blockstream:
Add basic profiling support to the nix definition.

@apoelstra you probably want to add this configuration to your CI system.  Feel free to test `withProfiler` together with `withCoverage`.  We do not need to CI every permutation here.

Also included in the PR is a ammendment to `test.c` that avoids running `computeAnnotatedMerkleRoot` when the result will not be used.  This lets us gather timing data that reflects production use.


apoelstra (approved):
ACK cd6557a1a2047595252df3dea1060e520fe89e22

PR #215, opened 2024-01-14
Make separate jet lib

uncomputable:
Only a subset of files is required for jet execution. Our goal is to port this subset for WASM.

rust-simplicity-sys doesn't call the local Makefile, but a target for jets will make it easier to track WASM compatibility.

# Current progress

I identified a closed subset of files that includes all jet implementations.

[rust-simplicity-sys](https://github.com/uncomputable/rust-simplicity/tree/only-jets) uses the same set of files and successfully runs on x86_64 and WASM.

# Limitations

There are no unit tests because the parsing of programs etc. is not included.

Elements transaction environments cannot be built because allocations were removed.

### DIFF
--- a/C/wrappedJets/elements/jets.c
+++ b/C/wrappedJets/elements/jets.c
@@ -1,4 +1,4 @@
-#include "primitive/elements/jets.h"
+#include "../../primitive/elements/jets.h"
 #include "../wrappers.h"
 
 WRAP_(version)
--- a/C/wrappedJets/wrappers.h
+++ b/C/wrappedJets/wrappers.h
@@ -3,17 +3,33 @@
 
 #include "simplicity_assert.h"
 
+/* In test suites, like the Haskell testsuite, we use test jets via these wrappers.
+ * During that testing we allocate a correctly sized frame for each jet,
+ * and as part of the testing we want to verify that jet writes to the end of that specifically allocated frame.
+ *
+ * However, other uses of these jets may call jets in which jets will only write to part of the frame.
+ * (this would happen, for instance, when executing jets in a bit machine context.)
+ *
+ * Therefore we only enable these assertions when JET_FRAME_TESTING is enabled.
+ */
+#ifdef JET_FRAME_TESTING
+#  define JET_FRAME_TESTING_FLAG 1
+#else
+#  define JET_FRAME_TESTING_FLAG 0
+#endif
+#define simplicity_jet_frame_assert(cond) do { if (JET_FRAME_TESTING_FLAG) { simplicity_assert(cond); } } while(0)
+
 #define COREWRAP_(jet)                                                                                                        \
 bool c_##jet(frameItem* dst, const frameItem* src) {                                                                          \
   bool result = jet(dst, *src, NULL);                                                                                         \
-  simplicity_debug_assert (!result || 0 == dst->offset);                                                                                       \
+  simplicity_jet_frame_assert(!result || 0 == dst->offset);                                                                   \
   return result;                                                                                                              \
 }
 
 #define WRAP_(jet)                                                                                                            \
 bool c_##jet(frameItem* dst, const frameItem* src, const txEnv* env) {                                                        \
   bool result = jet(dst, *src, env);                                                                                          \
-  simplicity_debug_assert (!result || 0 == dst->offset);                                                                                       \
+  simplicity_jet_frame_assert(!result || 0 == dst->offset);                                                                   \
   return result;                                                                                                              \
 }
 
--- a/C/wrappedJets/wrappers.h
+++ b/C/wrappedJets/wrappers.h
@@ -1,17 +1,19 @@
 #ifndef WRAPPEDJETS_WRAPPERS_H
 #define WRAPPEDJETS_WRAPPERS_H
 
+#include "simplicity_assert.h"
+
 #define COREWRAP_(jet)                                                                                                        \
 bool c_##jet(frameItem* dst, const frameItem* src) {                                                                          \
   bool result = jet(dst, *src, NULL);                                                                                         \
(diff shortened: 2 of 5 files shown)


apoelstra:
LGTM though needs to be tied into the nix buildscripts somehow in order for me to test it.

roconnor-blockstream:
I'm thinking of co-mingling the request for wrappers with this PR, so that we get a lib of wrapped jets, which use pointer arguments for everything.

uncomputable:
The changes in this PR are not used in the Rust FFI. Happy to convert this PR to draft until we know what we want from libsimplicity.

roconnor-blockstream:
Okay, this is my proposal.

@apoelstra The makefile will by default build the jets library.

It is untested as a static library.  I toyed with the idea of linking it with the Haskell code, but it seems just better to have Haskell compile the C code by itself thank to convince it to use a static library.  I suspect the same will be true with Rust bindings.

apoelstra:
Sounds good to me.

Re "just compiling the C code by itself" -- this does, in effect, create a static library. Just not one that was built through the Makefile.

C/wrappedJets/wrappers.h
```diff
@@ -1,5 +1,5 @@
-#ifndef WRAPPERS_H
-#define WRAPPERS_H
+#ifndef WRAPPEDJETS_WRAPPERS_H
+#define WRAPPEDJETS_WRAPPERS_H
 
 #define COREWRAP_(jet)                                                                                                        \
 bool c_##jet(frameItem* dst, const frameItem* src) {                                                                          \
```
roconnor-blockstream, line 5: The `assert`s in this file (below the fold) should be changed to some form of `simplicity_assert`s.

uncomputable:
Requiring a Haskell environment to build the C library would be a mess. Let's keep the in-Haskell-generated C files.

apoelstra (approved):
ACK 34fe6f9127d1e22570a8d904672905b628fbaf8b

C/wrappedJets/wrappers.h
```diff
@@ -1,17 +1,19 @@
...
+#ifndef WRAPPEDJETS_WRAPPERS_H
+#define WRAPPEDJETS_WRAPPERS_H
+
+#include "simplicity_assert.h"
 
 #define COREWRAP_(jet)                                                                                                        \
 bool c_##jet(frameItem* dst, const frameItem* src) {                                                                          \
   bool result = jet(dst, *src, NULL);                                                                                         \
-  assert (!result || 0 == dst->offset);                                                                                       \
+  simplicity_assert (!result || 0 == dst->offset);                                                                                       \
```
roconnor-blockstream, line 9: let's make these `simplicity_debug_assert`s.

roconnor-blockstream (changes requested):

apoelstra (approved):
ACK 9d2bda842240c3243faddc91752216cea0db5e99

C/wrappedJets/wrappers.h
```diff
@@ -1,17 +1,35 @@
-#ifndef WRAPPERS_H
-#define WRAPPERS_H
+#ifndef WRAPPEDJETS_WRAPPERS_H
+#define WRAPPEDJETS_WRAPPERS_H
+
+#include "simplicity_assert.h"
```
uncomputable, line 4:
```suggestion
#include "../simplicity_assert.h"
```
Doesn't compile for me otherwise.

roconnor-blockstream:
Upon review, this assert needed a bit more care.  The assert is only being used specifically in the context of Haskell testing of jets, where a (Simplicity) frame is allocated of exactly the right size to fit the jet output.  In other contexts the conditions of the assertion might be fine.

C/wrappedJets/wrappers.h
```diff
@@ -1,17 +1,35 @@
...
 #define COREWRAP_(jet)                                                                                                        \
 bool c_##jet(frameItem* dst, const frameItem* src) {                                                                          \
   bool result = jet(dst, *src, NULL);                                                                                         \
-  assert (!result || 0 == dst->offset);                                                                                       \
+  simplicity_jet_frame_assert(!result || 0 == dst->offset);                                                                   \
   return result;                                                                                                              \
 }
 
 #define WRAP_(jet)                                                                                                            \
 bool c_##jet(frameItem* dst, const frameItem* src, const txEnv* env) {                                                        \
```
roconnor-blockstream, line 30: Maybe we need to forward declare `txEnv` here?

C/wrappedJets/elements/jets.c
```diff
@@ -1,4 +1,4 @@
-#include "primitive/elements/jets.h"
+#include "../../primitive/elements/jets.h"
 #include "../wrappers.h"
 
 WRAP_(version)
```
uncomputable, line 4: > `COREWRAP_(calculate_issuance_entropy)`

Why are certain Elements jets wrapped without environment? Are those precisely the jets that don't read from the environment?

On the Rust side we always pass the environment to the jet. There is no way of knowing which jets read and which don't read from the environment. If it's not too much work on the Haskell side, let's wrap all Elements jets using `WRAP_`.

roconnor-blockstream:
I think it is time to abort this PR as it is.  The Haskell code just doesn't do what we want.  It has special `core` jet wrappers that allow jets to be tested with no environment.  The Haskell code has special asserts are that valid for its use case, but that do not hold in general.

Instead we should go back to making a second, similar set of wrappers for use in Rust.  Those wrappers can either be added to the Rust project or the C project, whichever you think is best.  If they are added to the C project we can place them in the jet library.

uncomputable:
I close this PR because Haskell and Rust need different sets of jet wrappers.

A better approach is to generate Rust jet wrappers inside the FFI using the Haskell code generators in #218.

PR #218, opened 2024-01-30
Generate Rust for WASM

uncomputable:
Update the Rust code generator to bind wrapped jet functions (which take source frames by reference). Raw jet functions (which take source frames by value) are invisible to the Rust code, which is needed for WASM. Add a generator of wrapped jet functions.

### DIFF
--- a/Haskell-Generate/GenRustJets.hs
+++ b/Haskell-Generate/GenRustJets.hs
@@ -370,12 +370,19 @@ rustFFIImports = vsep (map (<> semi)
 
 rustFFISigs :: Module -> Doc a
 rustFFISigs mod = vsep
-  [ nest 4 (vsep ("extern \"C\" {": ((<> semi) . pretty . sig <$> moduleJets mod)))
+  [ nest 4 $ vsep $
+    "extern \"C\" {" :
+    (declaration <$> moduleJets mod)
   , "}"
   ]
  where
-  sig (SomeArrow jet) = "pub fn "++cJetName jet++"(dst: *mut CFrameItem, src: CFrameItem, env: *const "++envType++") -> bool"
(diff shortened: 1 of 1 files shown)


apoelstra:
Did you forget to commit a new `wrapper.h` file? I don't see where this is.

Can you also say a few words about your strategy here -- it looks like you're renaming all the jet functions to add `_wrapper` to their names?

apoelstra:
Ahh, the file here is called `wrappers.h` but in the Rust world you call it `wrapper.h`.

Let's use the same name and just `cp` it in our vendoring script rather than modifying it.

uncomputable:
Sure, we can copy `wrappers.h` from Haskell cbits.

I updated the Rust FFI to use the original jet names, with the wrapper names as link name. The wrapper name is updated to match `wrappers.h`.

roconnor-blockstream:
I could probably move the wrapper from the `Haskell` directory to the `C` if that would be helpful.

apoelstra:
@roconnor-blockstream I think that's wise, now that two different language bindings separately found the need for it.

uncomputable:
Changes in https://github.com/BlockstreamResearch/rust-simplicity/pull/204 may change the import paths of types in the generated files. Converting to draft.

uncomputable:
Let's wait for #215 to be merged. Changes in the wrapper macros changed the signature of Core jets, which no longer include the environment parameter.

uncomputable:
Rebased and ready for review @apoelstra 

We're back to using wrapper.h from the Rust FFI. https://github.com/BlockstreamResearch/rust-simplicity/pull/208 is ready to use the updated Haskell generators.

apoelstra (approved):
ACK 9a1d681427b8b5384bafd758d0160c62c4344a8e

PR #222, opened 2024-02-13
Simplicity alloc

uncomputable:
Simplicity (de)allocation macros in the style of `simplicity_assert.h`. These macros can be overwritten to use Rust functions that are injected into C via FFI.

Environment allocation functions that return the number of allocated bytes.

### DIFF
--- a/C/eval.c
+++ b/C/eval.c
@@ -1,9 +1,9 @@
 #include "eval.h"
 
 #include <string.h>
-#include <stdlib.h>
 #include "bounded.h"
 #include "limitations.h"
+#include "simplicity_alloc.h"
 #include "simplicity_assert.h"
 
 /* We choose an unusual representation for frames of the Bit Machine.
@@ -589,7 +589,7 @@ simplicity_err analyseBounds( ubounded *cellsBound, ubounded *UWORDBound, ubound
   static_assert(DAG_LEN_MAX - 1 <= UINT32_MAX, "bound array index does not fit in uint32_t.");
   simplicity_assert(1 <= len);
   simplicity_assert(len <= DAG_LEN_MAX);
-  boundsAnalysis* bound = malloc(len * sizeof(boundsAnalysis));
+  boundsAnalysis* bound = simplicity_malloc(len * sizeof(boundsAnalysis));
   if (!bound) return SIMPLICITY_ERR_MALLOC;
 
   /* Sum up the total costs.
@@ -744,7 +744,7 @@ simplicity_err analyseBounds( ubounded *cellsBound, ubounded *UWORDBound, ubound
     *frameBound = bound[len-1].extraFrameBound[0] + 2; /* add the initial input and output frames to the count. */
     *costBound = bound[len-1].cost;
   }
-  free(bound);
+  simplicity_free(bound);
   /* Note that the cellsBound and costBound computations have been clipped at UBOUNDED_MAX.
    * Therefore setting maxCells or maxCost to UBOUNDED_MAX will disable the corresponding error check.
    */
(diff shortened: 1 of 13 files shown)


roconnor-blockstream:
In C, malloc always aligns allocation with respect to `alignof(max_align_t)`.  Should we just use that everywhere?

apoelstra:
I think we should. It would de-noise the C code and remove risk of copy/paste bugs that the compiler won't help us with.

We would need to export the size/align of `max_align_t` though, just like we do with a bunch of other C types, so that the Rust side can sanity-check its own values.

uncomputable:
Thanks for pointing this out. I removed the alignment parameter from the function signatures.

roconnor-blockstream:
>  just allocate extra memory and store the size in there so that we can extract it to deallocate.

I think this would be better.

roconnor-blockstream:
Be wary of alignment issues though.

C/simplicity_alloc.h
```diff
@@ -0,0 +1,34 @@
+#ifndef SIMPLICITY_SIMPLICITY_ALLOC_H
+#define SIMPLICITY_SIMPLICITY_ALLOC_H
+
+#include <stdlib.h>
+
+/* Allocate 'size' many bytes
```
roconnor-blockstream, line 6: We don't need these comments on the C code.  You can have just one comment that say something like "instantiating simplicity_malloc and simplicity_free with malloc and free from the stdlib.

uncomputable: I removed the comments.

uncomputable:
Allocation is done in Rust in this PR: https://github.com/BlockstreamResearch/rust-simplicity/pull/205

C/simplicity_alloc.h
```diff
@@ -0,0 +1,12 @@
+#ifndef SIMPLICITY_SIMPLICITY_ALLOC_H
+#define SIMPLICITY_SIMPLICITY_ALLOC_H
+
+#include <stdlib.h>
+
+/* Allocate with malloc by default. */
+#define simplicity_malloc(size) malloc(size)
```
roconnor-blockstream, line 7: This can be simply

    #define simplicity_malloc malloc

and similarly for free

roconnor-blockstream (changes requested):
Might as well rebase while you are at it.

uncomputable:
Rebased and compressed the macro definitions.

roconnor-blockstream:
@apoelstra I'm not sure if we should go ahead and replace all instances of malloc and free, or just these ones.  Do you have any thoughts?

uncomputable:
The original signature of `simplicity_alloc` was different from `malloc`, so only some files were changed. Now that the signatures are the same, it should be easy to use `simplicity_alloc` everywhere.

apoelstra:
> @apoelstra I'm not sure if we should go ahead and replace all instances of malloc and free, or just these ones. Do you have any thoughts? Having a mix of `simplicity_malloc` and `malloc` seems potentially dangerous ... maybe?

I think we should replace them all.

I *think* it's safe to mix in our current setup, where if you try to call the norrmal `malloc` in WASM things just won't compile, and otherwise they're the same. But this feels a bit fragile.

uncomputable:
I replaced all occurrences of `malloc` and of `free`.

I kept `calloc` in test.c because it is independent. For the record, we can define `simplicity_malloc` as `NULL` and the `make check` succeeds.

apoelstra (approved):
ACK b5492dab6c927757e2336216c5cf7c19b680afe3

C/eval.c
```diff
@@ -806,7 +806,7 @@ simplicity_err evalTCOExpression( flags_type anti_dos_checks, UWORD* output, con
   static_assert(2*DAG_LEN_MAX <= SIZE_MAX / sizeof(frameItem), "frames array does not fit in size_t.");
```
roconnor-blockstream, line 806: There is a `calloc` in the line above, which is a problem, because we definitely shouldn't be mixing `simplicity_free(cell)` with `calloc(cells)`

roconnor-blockstream:
Okay so I didn't think about the fact that we use `calloc`.

I can see two choices, either we add `calloc` to the `simplicity_` allocation interface, or we carefully divide our uses of memory allocation into internal and external uses?

But if our goal is to make the entire simplicity execution library (excluding test.c) as optionally independent from `stdlib.h` then maybe our only choice to to add `calloc` to the interface.

C/eval.c
roconnor-blockstream: Seems there is a second `calloc` in this file as well.

uncomputable: I replaced all occurrences of `calloc`. I included the occurrences in test.c for consistency.

apoelstra (approved):
ACK 683b6534235821f70667bba1f49d2f2799d9f9e0

roconnor-blockstream:
LGTM.  Can you squash the commits so we don't have any commits with unmatched `simplicity_free` `*alloc` floating around?

roconnor-blockstream:
@apoelstra when you get a chance can you rerun CI?

apoelstra (approved):
ACK adb7be91acd1e91415b93a5b29c634c4447f7a50

PR #224, opened 2024-03-08
Per input/output transaction hash jets

roconnor-blockstream:
The `tx-hash` and `*-hash` jets for Elements transactions generally hash parts of the transaction across all input and or outputs. This lets you do something like CTV and APO. However, for covenants you typically want to hash all the data that corresponds to a specific input or specific output.

* `input-hash : TWO^32 ⊢ S(TWO^256)` Similar to `inputs-hash` but only hashing the data of one input.
* `output-utxo-hash : TWO^32 ⊢ S(TWO^256)` Similar to `outputs-hash` but only hashing the data of one output.
* `issuance-hash : TWO^32 ⊢ S(TWO^256)` Similar to `issuances-hash` but only hashing the data of one input.
* `input-utxo-hash : TWO^32 ⊢ S(TWO^256)` Similar to `inputs-utxos-hash` but only hashing the data of one input.

This PR refactors some of the existing jet specifications changing their CMRs.


C/primitive/elements/jets.h
```diff
@@ -85,19 +85,23 @@ bool output_scripts_hash(frameItem* dst, frameItem src, const txEnv* env);
 bool output_range_proofs_hash(frameItem* dst, frameItem src, const txEnv* env);
 bool output_surjection_proofs_hash(frameItem* dst, frameItem src, const txEnv* env);
 bool outputs_hash(frameItem* dst, frameItem src, const txEnv* env);
+bool output_hash(frameItem* dst, frameItem src, const txEnv* env);
```
uncomputable, line 88: The code says `output_hash` but the PR description says `output-utxo-hash`. Let's update the latter.

C/primitive/elements/jets.h
```diff
@@ -85,19 +85,23 @@ bool output_scripts_hash(frameItem* dst, frameItem src, const txEnv* env);
 bool output_range_proofs_hash(frameItem* dst, frameItem src, const txEnv* env);
 bool output_surjection_proofs_hash(frameItem* dst, frameItem src, const txEnv* env);
```
uncomputable, line 86: Let's add `output_surjection_proof_hash`. The `~proofs_hash` is part of `tx_hash` so it seems we need the surjection proof hash to fully commit to an output.

roconnor-blockstream: The TL;DR is that surjection proofs depend on both the inputs and outputs and thus are not appropriate for output only sighash operations.

The longer story is that both the output range proofs and surjection proofs are not covered by the signature at all in legacy sighashes, since they are basically of fixed size and, as proofs, are not really supposed to have information in any information theoretic sense.  In element's taproot deployment the output range proofs and surjection proofs were added to the sighash in order to to cover all bases.  However we didn't think too much about this at the time and it has resulted in complaints about certain sighash modes.

As I noted above the surjection proofs depend on the inputs, and sometimes when users want to sign in some output only modes, the inputs needed in order to generate the surjection proofs are not yet available, making some tasks impossible to do with taproot based signatures due to the inclusion of surjection proofs in those sighashes.

roconnor-blockstream: This probably needs to be documented somewhere, but I'm not really sure where.  It is kinda of a quirky Element's thing.

uncomputable: The Elements' sighashes differ from Bitcoin's in interesting ways. I will write some documentation in my https://github.com/uncomputable/bitcoin-notes repo. We should add a comment to the output-surjection-proofs jet explaining why it doesn't make sense for single outputs.

Fair point that a ZKP is random noise from an information theoretic sense. The ZKP is malleable if it is not covered by the signature. However, it is exponentially unlikely that someone will fake the proof.

roconnor-blockstream:
@apoelstra This is ready for your review.

apoelstra:
LGTM.

Two observations:
* When hashing issuance data we also cover the rangeproof/surjectionproof of the issuance even though this is arguably witness data. Unsure if this is the right choice.
* When hashing inputs we don't hash the nonce of the spent output. I think this is correct, but in any case we have no choice, because Elements does not keep track of the nonce in the utxoset. (This is an old elements bug.)

apoelstra:
Looks like `primitiveEnumJet.inc` is not updated

```
++++ diff primitiveEnumJet.inc /nix/store/571z6kp50kw0x4v2x5vsgg3qk4qqp4jl-source/C/primitive/elements/primitiveEnumJet.inc
174a175
> INPUT_HASH,
183a185
> INPUT_UTXO_HASH,
201a204
> ISSUANCE_HASH,
329a333
> OUTPUT_HASH,
```

roconnor-blockstream:
Should be fixed now.

apoelstra (approved):
ACK 21dd38921bb42d19e5a7d354a8e20d8574340b70

PR #225, opened 2024-03-28
Add transaction_id jet

roconnor-blockstream:


Haskell/Core/Simplicity/Digest.hs
```diff
@@ -124,6 +124,10 @@ bslHash = ivHash . IV . pushChunks sha256Incremental . padSHA1
 bsHash :: BS.ByteString -> Hash256
 bsHash = bslHash . BSL.fromStrict
 
+-- | Computes a Bitcoin-style doulbe SHA-256 hash from a lazy 'BSL.ByteString'.
```
apoelstra, line 127: In 7f77a91d529ca7af00b2fbb8998a8cc641fd0010:

nit: typo `double`

apoelstra (approved):
ACK 389de8361f3bc9e8532a9bae9e303ac3ea51ec43

roconnor-blockstream:
> The risk I suppose is that in theory, the specification could fail to match the C++ code

By specification I usually mean the simplicity program that is being hashed to produce the CMR for the jet.  However in this case the "formal specification" is just `primitive Prim.TransactionId` which makes the CMR approximately the hash of the string "TransactionId". i.e. the isn't a formal specification, and the operation is considered primitive.

The Haskell code to compute the txid is just used to fill in the environment, which holds the 256 bit number that is used when this primitive is run.  If the Haskell code is wrong, then it just means our Haskell test vectors will have incorrect txids, which isn't great, but won't really have any consequences on the correctness of the C code.

The biggest risk here is when integrating with the elements codebase we need to make sure it is connected up correctly and, for example, the txid doesn't end up appearing byte-swapped or anything like that.

uncomputable:
Correct me if I'm wrong: The txid jet is a primitive that reads a value (the txid) from the environment. The specification of the primitive is to produce the exact value of the "txid" entry in the environment. The entry must exist and can hold any 256-bit number.

roconnor-blockstream:
>  The entry must exist and can hold any 256-bit number.

Technically it can hold any 256-bit number, but it would be improper for it to hold any value that isn't the txid.

PR #226, opened 2024-04-16
Lbtc assetid

roconnor-blockstream:
Jet that returns the asset id for l-btc on liquid v1.


Haskell/Elements/Simplicity/Elements/DataTypes.hs
```diff
@@ -689,3 +690,49 @@ outputFee txo = do
   Explicit assetId <- Just . view (under asset) $ txoAsset txo
   Explicit value <- Just . view (under amount) $ txoAmount txo
   return (assetId, value)
+
+-- | The (explicit) asset ID for liquid BTC.
+-- Note that this asset is specific to the production liquid blockchain.
+lBtcAsset :: Hash256
+lBtcAsset | result == review (over le256) 0x6f0279e9ed041c3d710a9f57d0c02928416460c4b722ae3457a11eec381c526d = result
```
uncomputable, line 697: I keep getting confused by the reversed asset ids. To reiterate, the C code and the Simplicity specification use the order where the LBTC id starts with `0x6d`. The Haskell code uses the order where the same id starts with `0x6f`. I think elementsd also returns `0x6f` first.

C/primitive/elements/primitiveJetNode.inc
```diff
@@ -1671,6 +1671,14 @@
 , .targetIx = ty_w256
 , .cost = 281 /* milli weight units */
 }
+,[LBTC_ASSET] =
+{ .tag = JET
+, .jet = lbtc_asset
+, .cmr = {{0x495dd0a1u, 0xcb39ee84u, 0xff45e1dfu, 0x3db4b69du, 0xe27c37c3u, 0x6d69b312u, 0x976c43f5u, 0x0e5d941au}}
+, .sourceIx = ty_u
+, .targetIx = ty_w256
+, .cost = 3065 /* milli weight units */
```
uncomputable, line 1680: The jet is surprisingly expensive, given that is writes a 256-bit constant. The current-asset jet costs 320 milli weight units.

uncomputable (approved):
ACK 515d537000d684931d06d07e267450d900bf278c The test suite runs and no unexpected changes to the code.

apoelstra (approved):
ACK 515d537000d684931d06d07e267450d900bf278c

Haskell/Elements/Simplicity/Elements/DataTypes.hs
roconnor-blockstream: Elementsd only *displays* asset ids with 0x6f first.  Internally both Simplicity and Elementsd being with 0x6d.

C/primitive/elements/primitiveJetNode.inc
roconnor-blockstream: We (and by we I mean you) will need to benchmark all the new jets next month once everything is ready.  For now I just cloned the cost of 'AnnexHash', which is looks up and prints a 256-bit value from the environment.

Haskell/Elements/Simplicity/Elements/DataTypes.hs
roconnor-blockstream: the `le` in `review (over le256)` stands for "little endian", meaning it parses the integer in little endian order to create the hash value.

roconnor-blockstream: But yes, part of the purpose of this code is to compute the hash value by actually hashing the input, thus guaranteeing that the byte order is internally correct (as the byte-reversed version would have no known preimage), and then validating that the hash value matches the expected value.  (And if the expected value were incorrectly written backwards, there would be no way to get this procedure to work).

C/primitive/elements/primitiveJetNode.inc
uncomputable: Got it, haha :laughing:

Haskell/Elements/Simplicity/Elements/DataTypes.hs
uncomputable: That makes sense. Big endian is what humans are used to seeing and what Simplicity uses. The Haskell code uses little endian because the literal bytes are processed by something else and the output would be different for big endian. elementsd usually prints values in little endian. Thanks for clarifying.

PR #227, opened 2024-04-26
DivMod128_64 jet

roconnor-blockstream:
This jet divides a 128 bit number A by a 64 bit number B, returning the quotient Q and remainder R under the conditions that 2^63 ≤ B and Q is 64 bits.

This jet isn't that useful directly, however it is the basis for recursively defining wide integer division functions.


C/jets.c
```diff
@@ -1043,6 +1043,165 @@ DIVIDES_(16)
...
+     *
+     * Then append the last "digit" of A to the intermidiate remainder and divide that value (96_bits) by the two "digit"s (64-bits) of B,
+     * returning the second "digit" (low 32-bits) of the quotient, and the final remainer consisiting of 2 "digit"s (64-bits).
+     */
+    div_mod_96_64(&ql, &r, r, al, b);
+    write32(dst, qh);
+    write32(dst, ql);
+    write64(dst, r);
+  } else {
+    /* Set all the bits in the output when the input is out of bounds. */
```
apoelstra, line 1198: In d80cc978a89103d5a5717a1b105a010d28d02e06:

Why did you choose to this rather than returning an optional type?

apoelstra: I guess, the idea is that this is a primitive and if you call this with wrong values that's a programming error, and we don't want to force the caller to destructure things in that case?

You could just `fail` in this case.

apoelstra (approved):
ACK d80cc978a89103d5a5717a1b105a010d28d02e06 nice! I like the long division wrapper.

C/jets.c
roconnor-blockstream: Yeah.  The main disadvantage with using `fail` is that the term is no longer in the Core (assertion free) subset.  It's hard to tell what is best here.

roconnor-blockstream: I don't go with an optional type because this is an expression to recursively build up division operators for wider and wider integers, and adding code to handle possible failure cases for inputs that one knows cannot fail is somewhat annoying.

Also worth point out that the remainder value returned is always larger than than what can be returned during a successful run.

PR #228, opened 2024-05-01
Refactor Elements Catalogue

roconnor-blockstream:
Now that there are no longer any missing items, we can use the `book` construction.

This ought to have been done in #177.


roconnor-blockstream:
@apoelstra Can you run this through your CI?

apoelstra:
Sorry, was at a conference and assumed this would be more work than it was. I wish github notifications would show me the total diff size.

apoelstra (approved):
ACK 0883b14c63312052b73165a40608afcdff480e8f

PR #230, opened 2024-05-10
GenRust: Update c_jet_env

uncomputable:
This function converts a Rust environment into a C environment. The specific jet is not used at all and can be dropped from the function signature.


apoelstra (approved):
ACK 3a5f9bdf3d8945bbf269e7c68ae47e321c3ba09b

uncomputable:
@roconnor-blockstream I changed the base of the PR to `staging`.

PR #231, opened 2024-05-22
Temporary work around for https://github.com/wrengr/unification-fd/issues/70

roconnor-blockstream:
unification-fd currently doesn't work on nixpkgs-unstable

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -21,6 +21,22 @@ let hp = nixpkgs.haskell.packages.${ghc};
     overrides = self: super: {
       Simplicity = haskell;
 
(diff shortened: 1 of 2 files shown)


roconnor-blockstream:
This change should be safe, but please make sure the Haskell simplicity project still builds on your stable channels for nixpkgs.

uncomputable (approved):
ACK 67817cedf9c1f42b8723fa64cc76c01a9137b857 There was more logging on the CLI than usual, but the Haskell project built and ran successfully.

apoelstra (approved):
ACK 67817cedf9c1f42b8723fa64cc76c01a9137b857

PR #233, opened 2024-06-05
Add API to compute CMRs

roconnor-blockstream:
Adding `simplicity_computeCmr` to compute the CMR of a given simplicity program. This is expected to be used in testing infrastructure, though it might also end up being used by wallet software.


apoelstra (approved):
ACK 847f6411d08791de38fce4c8983589c518a0b268

uncomputable (approved):
ACK 847f6411d08791de38fce4c8983589c518a0b268

PR #235, opened 2024-06-13
Separate witness data

roconnor-blockstream:
We will store witness data in a separate stack item.

Regression tests 1 and 2 used to test the prefix of the witness data which consisted of the length of the witness data.  However that prefix no longer exists.  Thus the tests have been removed.


roconnor-blockstream:
I have a personal branch of elements that integrates this interface change, just to make sure it works.  If you are interested you can find it at https://github.com/roconnor-blockstream/elements/tree/simplicity-staging.

I expect that https://github.com/uncomputable/asset-gen will need to be eventually reworked to work with this new interface to Simplicity.

apoelstra (approved):
ACK 4c77c31217a590cc933c0e27570704ec59c45cf1 looks good. will take some work to update rust-simplictiy but probably not too bad

C/dag.c
```diff
@@ -472,31 +472,31 @@ simplicity_err verifyCanonicalOrder(dag_node* dag, const size_t len) {
 /* This function fills in the 'WITNESS' nodes of a 'dag' with the data from 'witness'.
  * For each 'WITNESS' : A |- B expression in 'dag', the bits from the 'witness' bitstring are decoded in turn
  * to construct a compact representation of a witness value of type B.
- * This function only returns 'SIMPLICITY_NO_ERROR' when exactly 'witness.len' bits are consumed by all the 'dag's witness values.
- * If extra bits remain, then 'SIMPLICITY_ERR_WITNESS_TRAILING_BITS' is returned.
- * If there are not enough bits, then 'SIMPLICITY_ERR_WITNESS_EOF' is returned.
- *
- * Note: the 'witness' value is passed by copy because the implementation manipulates a local copy of the structure.
+ * If there are not enough bits, then 'SIMPLICITY_ERR_BITSTREAM_EOF' is returned.
```
uncomputable, line 475: 4c77c31217a590cc933c0e27570704ec59c45cf1 This documentation is wrong. `SIMPLICITY_ERR_WITNESS_EOF` is returned if there are not enough bits.

C/dag.h
```diff
@@ -364,17 +365,17 @@ simplicity_err verifyCanonicalOrder(dag_node* dag, const size_t len);
 /* This function fills in the 'WITNESS' nodes of a 'dag' with the data from 'witness'.
  * For each 'WITNESS' : A |- B expression in 'dag', the bits from the 'witness' bitstring are decoded in turn
  * to construct a compact representation of a witness value of type B.
- * This function only returns 'SIMPLICITY_NO_ERROR' when exactly 'witness.len' bits are consumed by all the 'dag's witness values.
- * If extra bits remain, then 'SIMPLICITY_ERR_WITNESS_TRAILING_BITS' is returned.
- * If there are not enough bits, then 'SIMPLICITY_ERR_WITNESS_EOF' is returned.
+ * If there are not enough bits, then 'SIMPLICITY_ERR_BITSTREAM_EOF' is returned.
```
uncomputable, line 368: 4c77c31217a590cc933c0e27570704ec59c45cf1: Should be `SIMPLICITY_ERR_WITNESS_EOF`

C/primitive/elements/exec.c
```diff
@@ -84,9 +82,15 @@ extern bool elements_simplicity_execSimplicity( simplicity_err* error, unsigned
...
-    }
+      bitstream witness_stream = initializeBitstream(witness, witness_len);
+      *error = fillWitnessData(dag, type_dag, (size_t)dag_len, &witness_stream);
       if (IS_OK(*error)) {
+        *error = closeBitstream(&witness_stream);
+        if (SIMPLICITY_ERR_BITSTREAM_TRAILING_BYTES == *error) *error = SIMPLICITY_ERR_WITNESS_TRAILING_BYTES;
+        if (SIMPLICITY_ERR_BITSTREAM_ILLEGAL_PADDING == *error) *error = SIMPLICITY_ERR_WITNESS_ILLEGAL_PADDING;
+      }
+    }
+    if (IS_OK(*error)) {
```
uncomputable, line 93: 4c77c31217a590cc933c0e27570704ec59c45cf1 If the witness is now a bitstream instead of a bitstring, then we could drop the specialized witness error variants in favor of the bitstream variants.

C/include/simplicity/errorCodes.h
```diff
@@ -67,6 +66,8 @@ static inline const char * SIMPLICITY_ERR_MSG(simplicity_err err) {
     return "Program has STOP node";
   case SIMPLICITY_ERR_HIDDEN:
     return "Program has illegal HIDDEN child node";
+  case SIMPLICITY_ERR_BITSTREAM_EOF:
+    return "Unexpected end of bitstream";
   case SIMPLICITY_ERR_BITSTREAM_TRAILING_BYTES:
```
uncomputable, line 71: If we unify the bitstream and witness error variants, then we have to update the error messages.

"Trailing bytes after final byte of witness" → "Trailing bytes after final byte of encoded object" (or something similar)

C/test.c
```diff
@@ -217,13 +217,17 @@ static void test_program(char* name, const unsigned char* program, size_t progra
...
         dag[len-1].sourceType != 0 || dag[len-1].targetType != 0) {
       failures++;
       printf("Unexpected failure of type inference.\n");
-    } else if (!IS_OK(fillWitnessData(dag, type_dag, (size_t)len, witness))) {
+    } else if (!IS_OK(fillWitnessData(dag, type_dag, (size_t)len, &witness_stream))) {
       failures++;
       printf("Unexpected failure of fillWitnessData.\n");
+    } else if (!IS_OK(closeBitstream(&witness_stream))) {
+      failures++;
+      printf("Unexpected failure clossing witness_stream\n");
```
uncomputable, line 230: Typo "clossing" → "closing"

uncomputable (changes requested):
4c77c31217a590cc933c0e27570704ec59c45cf1 Let's fix the documentation about which errors are returned upon invalid witness encodings. Optionally, let's remove the specialized witness error variants. The rest looks good.

C/primitive/elements/exec.c
roconnor-blockstream: I've thought about this, but I think it is more helpful to get an error message about which stack item had the wrong padding.

C/include/simplicity/errorCodes.h
roconnor-blockstream: What's wrong with `witness`?  Would you be happy with `Simplicity witness`?

C/primitive/elements/exec.c
uncomputable: Ok, in this case let's keep both variants. We could rename `SIMPLICITY_ERR_BITSTREAM_*` to `SIMPLICITY_ERR_PROGRAM_*` to clarify the distinction. If we do that, let's do that in a follow-up PR.

C/include/simplicity/errorCodes.h
uncomputable: If we keep program and witness error variants separate, then I'm happy with the error messages as they currently are.

roconnor-blockstream:
I've address the comments.

apoelstra (approved):
ACK 1b85dc31d80d36dc012755e4369aeeac815476a6

uncomputable (approved):
ACK 1b85dc31d80d36dc012755e4369aeeac815476a6

C/ctx8Pruned.c
```diff
@@ -260,6 +260,11 @@ const unsigned char ctx8Pruned[] = {
 };
 
 const size_t sizeof_ctx8Pruned = sizeof(ctx8Pruned);
+const unsigned char ctx8Pruned_witness[] = {
+
+};
+
+const size_t sizeof_ctx8Pruned_witness = sizeof(ctx8Pruned_witness);
```
apoelstra, line 267: In 1b85dc31d80d36dc012755e4369aeeac815476a6:

Over in rust-simplicity, our MSVC Windows CI seems to shit the bed on this. It issues a warning that `sizeof` returns 0 and then has an internal compiler error and crashes.

Is this legal C?

roconnor-blockstream: I think it is not legal.

roconnor-blockstream: Fixed in #246.

Issue #234, opened 2024-06-12
C: clamping of `budget` in `elements_simplicity_execSimplicity` is incorrect


apoelstra:
In `elements_simplicity_execSimplicity` we clamp the passed `budget` value to ensure it does not exceed `BUDGET_MAX`. But we do not check for negative values. If you pass a negative value, it will be casted to a `ubounded` that exceeds `BUDGET_MAX`, eventually triggering an assertion failure in eval.c.

There are three ways we might handle negative values:
* clamping to 0
* returning an error
* asserting in execSimplicity

I have no preference about which.

roconnor-blockstream:
I'm leaning towards making the budget non-negative a precondition and adding an assertion in execSimplicity.

roconnor-blockstream:
Fixed in 'staging'.

PR #236, opened 2024-06-18
Assert preconditions for execSimplicity

roconnor-blockstream:
Currently a `false` result in execSimplicity is ambigous. It either means a precondition failed, or it failed to complete its computation due to lack of memory or some other "transient" error.

But failing a preconditon is not "transient".  Instead we will just assert the various precondtions hold and require that the caller ensure that the preconditions are satisfied.

We also add a new 'non-negative budget' precondition that was missing before.

### DIFF
--- a/C/primitive/elements/exec.c
+++ b/C/primitive/elements/exec.c
@@ -30,6 +30,7 @@
  *               NULL != tx;
  *               NULL != taproot;
  *               unsigned char genesisBlockHash[32]
+ *               0 <= budget;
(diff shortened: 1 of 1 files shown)


roconnor-blockstream:
Fixes #234.

apoelstra (approved):
ACK 07c13677e9faa63a0c12faa286c8687d29869fed

uncomputable (approved):
ACK 07c13677e9faa63a0c12faa286c8687d29869fed

PR #237, opened 2024-06-20
Safer ecmult jets

roconnor-blockstream:
To be extra safe, ensure all jets that run libsecp256k1's ecmult implementation only operate on on-curve points.


apoelstra (approved):
ACK 376beea9c16b3081c6f3b0e2e5568fad4b985676

Issue #209, opened 2023-12-08
Cheap but slow program


uncomputable:
I found a program that seems to exceed the expected runtime given its cost.

The program is constructed as follows:
```
f(0)     := iden
f(n + 1) := comp (take f(n)) (drop f(n))
```
Importantly, the type signature of `f(n)` differs depending on how it is used. Sharing happens after the program is constructed.

Let's build a C test vector that measures the runtime. The Elements test suite is inaccurate because it runs the same program multiple times.

roconnor-blockstream:
This is pretty significant.

Profiling (see #214) shows that upwards of 60% of time is spent on computing various Merkle roots, in this kind of case.  The use of Merkle trees for syntax is what enables pruning and helps with sharing, but we do have to run at least one SHA256 compression function per node when computing a Merkle Root.  In this example there is slightly more that one node per serialized byte (i.e. per WU).  For each node, not only do we need to perform a CMR computation but we need to do an IMR computation as well, which needs to hash not only the node, but the input and output Type Merkle roots.  Additionally type Merkle roots themselves need to be computed, which again is potentially on the order of the size of the DAG, as it is in this test case.

The coverage analysis tool tells that we do can do up to 4 calls to the compression function per node (which is approximately per WU).  If we estimate the cost of the compression function to be 225 ns, and if we want to maintain our goal of doing no more than 500 ns of work per WU, then we are in trouble because we can only do 2 calls to the compression function per node, and that leaves basically no time to do any other of the type inference work.

A couple of options on how to deal with this issue come to mind:

### 1. We can presume/require that nodes have SHA2 hardware acceleration
@apoelstra suggests hashing would be 100x faster with hardware acceleration.  I think the improvements would be more modest than that, but it would certainly be faster.  We could benchmark with this extended instruction set and see if that solves the problem.

### 2. We could eat into the budget
We currently use a budget mechanism to require a certain amount of witness weight to cover the cost of evaluation (which can be exponential in the size of the program).  Stuffing the annex with empty data may be necessary to cover the cost of evaluation.  (Technically tapscript does the same budgeting; its just that the annex trick isn't required for anything but the most obscene programs.)

We could extend this budget mechanism to count the number of nodes and cover the cost of all the hashing using the same budgeting system.

### 3. We could revamp the entire serialization format

This solution is my current favourite.

I devised the current bit-based serialization format to try to squeeze Simplicity programs to be as small as possible.  But this issue suggests that the current serialization format is "too good".

[...]

PR #240, opened 2024-06-26
Sha ni

roconnor-blockstream:
Relax timing limit somewhat, and check for sha_ni CPU support.

Timing in testing require sha_ni CPU support.


C/cpu.c
```diff
@@ -0,0 +1,191 @@
+#include "simplicity/cpu.h"
+
+#if defined(__SHA__) && defined(__SSE4_1__) && (defined(__x86_64__) || defined(__amd64__) || defined(__i386__))
+#define SHA_NI
+#endif
+
+#ifdef SHA_NI
+
+#include <cpuid.h>
```
uncomputable, line 9: 3563f344d22b59ce27af5c07ff7998259d742699: If any header is undefined on the target architecture then there will be an error, right? It seems that x86_64, amd64 and i386 all support cpuid, etc.

C/regression4.h
```diff
@@ -0,0 +1,19 @@
+#ifndef SIMPLICITY_REGRESSION4_H
+#define SIMPLICITY_REGRESSION4_H
+
+#include <stddef.h>
+#include <stdint.h>
+#include "bounded.h"
+
+/* A length-prefixed encoding of the following Simplicity program:
+ *     uWitness OneV : f 15 ++ [uComp (3*2^16) 1]
```
uncomputable, line 9: This looks like the program in the OP of https://github.com/BlockstreamResearch/simplicity/issues/209, but with an additional `++ [uComp (3*2^16) 1]`?

uncomputable (approved):
ACK dceeaae127d9ab461ee470febeaae765c3cb0c9a Looks good and works on my end. I'm happy that these 9000 lines of changes stem from a single byte blob test vector. I was scared initially.

C/cpu.c
```diff
@@ -0,0 +1,191 @@
...
+#endif
+
+/* WARNING: This function isn't threadsafe and must not be called concurrently with any function in this library, including itself.
+ * It is encouranged that you call this function once in the main function of your program
+ *
+ * This function inspects the CPU's capabilities and enables the use of some optimized instructions, if they are known and available.
+ *
+ * If the CPU supports Intel sha_ni operations, the result value has the SIMPLICITY_HAS_SHA_NI_FLAG bit set.
+ */
+uint32_t simplicity_cpu_optimize_not_thread_safe(void) {
```
uncomputable, line 182: AFAICS, SHA NI is not enabled by default if it is available. One has to call the `simplicity_cpu_optimize_not_thread_safe` function to explicitly opt into the optimization.

C/cpu.c
roconnor-blockstream: `cpuid.h` is provided by gcc and clang.  It may just fail to exist on other compilers.

apoelstra:
utACK dceeaae127d9ab461ee470febeaae765c3cb0c9a. Will hold off on running my test rig while you are playing with stuff.

roconnor-blockstream:
So I f'd up the source branch of this PR.  I've made a new PR at #241.

I guess new comments should go there.

C/regression4.h
roconnor-blockstream: the last uComp composes the main program with `uWitness`.

Here the `:` is list cons in Haskell, which always confuses me when I read this comment to myself.

PR #244, opened 2024-07-11
Add tests and checks for type-skip scenarios

roconnor-blockstream:
Type-skip is an implementation feature that skips over deeply nested products of units (which have bitSize 0), in order to avoid exponential type traversals without any associated cost.

Previously this feature was not so well tested.  This testcase increases coverage (i.e. nix building with `--arg withCoverage true`) in the `type.h`, `dag.c`, and `eval.c` files.  We also add some related assertions.


C/typeSkipTest.c
```diff
@@ -0,0 +1,62 @@
...
+ *       m8 = take l3 &&& drop m7
+ *       n1 = copair l3 l3
+ *       n2 = take n1 &&& drop l1
+ *       n3 = take l2 &&& drop n2
+ *       n4 = take n3 &&& drop l3
+ *       n5 = copair (injl ltop) (injr n4)
+ *       n6 = take n5 &&& drop l0
+ *       n7 = take l1 &&& drop n6
+ *       n8 = take n7 &&& drop l2
+ *       mn = copair (injl m8) (injr n8)
```
uncomputable, line 26: This looks like an optimized version of a [similar test case in asset-gen](https://github.com/uncomputable/asset-gen/blob/aad0f90ad7ff066c4ab54bb2015767e4c21939b9/src/main.rs#L46). Out of curiosity, could you explain what the program in this PR does that the program in asset-gen doesn't do?

uncomputable (approved):
ACK ce60b956a723b5d58dcf792f0976cb6c4b005c87

C/typeSkipTest.c
roconnor-blockstream: I consider this test different than the one in asset-gen.

The one in asset-gen is trying to illustrate exponential time savings by giving a case that runs fast when type skip is implemented and would otherwise run slow.

The test here is trying to extend code coverage by hitting various branches in `type.h`'s `typeSkip`, `dag.c`'s `fillWitnessData`, and `eval.c`'s `writeValue`.

apoelstra (approved):
ACK ce60b956a723b5d58dcf792f0976cb6c4b005c87 will try to remember to pull this new test vector into rust-simplicity

PR #245, opened 2024-07-12
Fix bindings to scale and linear_combination_1

roconnor-blockstream:
When the safe_ versions of these functions were introduced, the primitiveJetNode.inc file was not correctly updated to bind to the new "safe_" version of the jets.

To remedy this issue we instead exchange the names of the safe_ functions, removing the safe_ prefix and instead adding an "off_curve_" prefix to the unsafe versions.

primitiveJetNode.inc isn't touched in this commit since it was already referencing the names without a prefix.


roconnor-blockstream:
This fixes an issue with #237.

C/jets-secp256k1.c
```diff
@@ -500,7 +500,7 @@ bool ge_is_on_curve(frameItem* dst, frameItem src, const txEnv* env) {
   return true;
 }
 
-bool scale(frameItem* dst, frameItem src, const txEnv* env) {
+bool off_curve_scale(frameItem* dst, frameItem src, const txEnv* env) {
```
uncomputable, line 503: Before, the "safe" versions were safe and the normal unprefixed versions were unsafe. With this PR, the "unsafe" versions are unsafe and the normal unprefixed versions are safe. This encourages the use of the safe versions. Is this correct?

uncomputable:
@roconnor-blockstream Do you plan to update primitiveJetNode.inc to include `off_curve_scale` and `off_curve_linear_combination1`?

apoelstra (approved):
ACK 2b68236ee0901386b9c12c8e614f5ab170adb25c

uncomputable (approved):
ACK 2b68236ee0901386b9c12c8e614f5ab170adb25c

PR #246, opened 2024-07-16
Fixup empty arrays in test vectors

roconnor-blockstream:
Arrays cannot be of zero size in C.

Normally I would change them to NULL pointers, however one cannot assign a NULL value to an array.  I could, and maybe should, change the header so that these values are pointers and not array, but if I were to do that, I ought to change all of the types of the test vectors to be pointers, in order to be consistent. In which case to assign an array to the pointer I'd have to use something like a compound literal.

That seems like a big pain for just test vectors, so for now we are just going to put a null character into what should be empty arrays and explicity set their corresponding size value to 0.

### DIFF
--- a/C/ctx8Pruned.c
+++ b/C/ctx8Pruned.c
@@ -260,11 +260,10 @@ const unsigned char ctx8Pruned[] = {
 };
 
 const size_t sizeof_ctx8Pruned = sizeof(ctx8Pruned);
-const unsigned char ctx8Pruned_witness[] = {
+const unsigned char ctx8Pruned_witness[] = "";
 
-};
 
-const size_t sizeof_ctx8Pruned_witness = sizeof(ctx8Pruned_witness);
+const size_t sizeof_ctx8Pruned_witness = 0;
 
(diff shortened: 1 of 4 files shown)


apoelstra (approved):
ACK 714b44dafd66ab5d164c9247a0f793c320272162 though I think [0] would be a little clearer than ""

roconnor-blockstream:
> … I think [0] would be a little clearer than ""

I slightly prefer '""' in order to syntactically distinguish it from the deliberate single byte array '{ 0x00 }'.

uncomputable:
@apoelstra Have you tested if this fix makes the Windows compiler happy? It's kind of hard to check without running Windows.

apoelstra:
@uncomputable no, I haven't checked. But it will certainly fix the invalid C and the "zero sized array" warning, which we should really fix even if we didn't have a compiler whining at us.

If there are other instances it's probably easiest to just do another PR. And if MSVC is ICEing for unrelated reasons ... well, I hope that's not the case.

PR #248, opened 2024-07-24
Ensure that all symbols in the simplicity library begin with `simplicity_`

roconnor-blockstream:
C has no name spaces, and we want to ensure there are no symbol collisions with other libraries.

You can check the list of exported symbols by somelike

    nm -C result/lib/libElementsSimplicity.a | grep " [A-TV-Z] "

The `grep` is largely to exclude non-exported symbols from the list.


roconnor-blockstream:
In rust-simplicty you will want to ammend

https://github.com/BlockstreamResearch/rust-simplicity/blob/c74f94dd9565a750abe47e3ea51045eea5793d0b/simplicity-sys/depend/wrapper.h#L6

to 

      bool result = simplicity_##jet(dst, *src, env); \

uncomputable (approved):
ACK d2cdd728252458b22703645db1179f68707845f7 It's sad that we have to do this, but it seems common practice for C libraries :/

apoelstra:
The 5184676a4523b3054da2cad2f11c9179352e7460 `haskell` derivation doesn't compile. Is your goal that every individual commit of this PR will build?

roconnor-blockstream:
Might be working now.

apoelstra (approved):
ACK bd89d2a2bca5f008812f0eac68d9b84fbf2a1a83

PR #249, opened 2024-08-07
Fix some technical C typing issues

roconnor-blockstream:
This PR fixes up the types and makes for error free compiling on 32-bit and other exotic systems.

Most of the commits don't even make change in the binary generated on 64-bit systems, but the last commit, which unfortunately also is the most complicated one, does change the binary by a little bit.


apoelstra (approved):
ACK 84af1aca6a2a1d692fffcf3fe36507da27ad93f3 successfully ran local tests; FWIW it seems like many of these new casts actually prevent integer promotion, though I suppose in non-observable ways

PR #250, opened 2024-08-08
Staging

roconnor-blockstream:
Merge staging back into master


roconnor-blockstream:
The big question is, do we want to do a merge, or do we want to do a rebase?  Rebasing will change all the commit ids that we carefully acked.  Merging might mess up the C-master subtree branch, but probably not significantly, if at all.

roconnor-blockstream:
git subtree split -P C -b C-master

seems to work just fine after merging.

apoelstra:
I say we merge.

ACK 84af1aca6a2a1d692fffcf3fe36507da27ad93f3 which I already ACKed in #249.

PR #251, opened 2024-08-22
Rust: Generate array of all jets

uncomputable:
Generate an array of all jets because there is no way to access this array from raw Rust enums.

### DIFF
--- a/Haskell-Generate/GenRustJets.hs
+++ b/Haskell-Generate/GenRustJets.hs
@@ -286,10 +286,21 @@ rustJetImpl mod = vsep $
 
 rustJetEnum :: Module -> Doc a
 rustJetEnum mod = vsep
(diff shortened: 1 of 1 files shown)


roconnor-blockstream:
LGTM.  @apoelstra can you run this through your CI?

apoelstra:
I think we should add a newline between each element of the output. Right now it outputs a 469-element array all on one line, which is fine (total length is under 10k so it shouldn't break any editors) but looks a little ugly.

uncomputable:
I was wondering if we should make the array horizontal or vertical. I can make it vertical if you want.

uncomputable:
I made the array vertical.

apoelstra (approved):
ACK 3daad68dc9e995fefcd5287fa5b15a7bfb13462a successfully ran local tests

PR #257, opened 2024-09-17
List C sources

roconnor-blockstream:
This can be used for automake tooling, such as in Elements.


apoelstra (approved):
ACK 472b2105eacf00eedda0d9cc7c97c9c5edd9e9da successfully ran local tests

PR #259, opened 2024-09-19
Suppress signed/unsigned comparision warnings from MSVC

roconnor-blockstream:
This is probably due to signed integer promotion, so ensuring both sides are promoted to at least unsigned integer size removes the warning.


uncomputable (approved):
ACK 2dd04c757587e35f395dfe404f877b3a05234189

apoelstra (approved):
ACK 2dd04c757587e35f395dfe404f877b3a05234189 successfully ran local tests

PR #260, opened 2024-09-20
Msvc jets.c

roconnor-blockstream:
Renaming files to make MSVC happier, and renaming sources.mk at the same time to make it more clear that it is elements specific.


roconnor-blockstream:
This renaming might require changes in rust-simplicity.

apoelstra:
I don't think it will; we vendor only the contets of the C/ directory.

apoelstra:
```

error: builder for '/nix/store/0gp4zpjxcpm4dqqik09idiqllbm03krj-Simplicity-0.0.0.drv' failed with exit code 1;
       last 10 log lines:
       > provides this library (you may need the "-dev" version). If the library is
       > already installed but in a non-standard location then you can use the flags
       > --extra-include-dirs= and --extra-lib-dirs= to specify where it is.If the
       > library file does exist, it may contain errors that are caught by the C
       > compiler at the preprocessing stage. In this case you can re-run configure
       > with the verbosity flag -v3 to see the error messages.
       > If the header file does exist, it may contain errors that are caught by the C
       > compiler at the preprocessing stage. In this case you can re-run configure
       > with the verbosity flag -v3 to see the error messages.
       >
       For full logs, run 'nix log /nix/store/0gp4zpjxcpm4dqqik09idiqllbm03krj-Simplicity-0.0.0.drv'.
```

Haskell derivation fails to build with this patch.

Also, if you are changing it anyway, I would suggest putting a `-` in the name.

roconnor-blockstream:
... also this renaming might require changes in the haskell bindings.

roconnor-blockstream:
I think it is fixed.

apoelstra:
Oh, I see. Yeah, this likely will require changes in rust-simplicity.

apoelstra (approved):
ACK 03d38d1759316d7134c7b9bd6c687d2e2d1423ca successfully ran local tests

PR #262, opened 2024-10-01
Remove 'Draft' from the Simplicity Prefix

roconnor-blockstream:


uncomputable (approved):
7592dd576d32235cc98c4abc4ad400a0f218a244 ACK Looks good to me; tests run

apoelstra (approved):
ACK 7592dd576d32235cc98c4abc4ad400a0f218a244 successfully ran local tests

C/prefix.h
```diff
@@ -1,3 +1,3 @@
 #ifndef SIMPLICITY_PREFIX
-#define SIMPLICITY_PREFIX "Simplicity-Draft"
```
roconnor-blockstream, line 2: Oops, this prefix is incorrect.  It should be "Simplicity" not "Simplicity-1" (something that I was toying with in an earlier revision).

I'm surprised this didn't tickle any test error, but it turns out that since doing `precomputation.h`, this isn't actually used anymore.  I'll followup by simply deleting this.

PR #264, opened 2024-10-02
Remove unused prefix.h

roconnor-blockstream:
Since about the time of b2b196b23a54120227d96dfe39d83211cdb9a4c4 the SIMPLICITY_PREFIX has actually been unused, which explains why when 7592dd576d32235cc98c4abc4ad400a0f218a244 was merged containing an error, it didn't cause any test failures.

So here we fix the issue by simply deleting the file.

Fixes a minor issue with [#262](https://github.com/BlockstreamResearch/simplicity/pull/262).


apoelstra (approved):
ACK aba75aefc53982253cd1c8b3819e2a93db932c11 successfully ran local tests

PR #272, opened 2024-11-08
Proof of correctness for secp256k1_modinv64_var

roconnor-blockstream:


roconnor-blockstream:
This is ready now.  You should be able to run CI on all commits.

apoelstra (approved):
ACK c1dddedd553b403da877377e658f17f0d2184cc4; successfully ran local tests

uncomputable:
Casually merging 300000 lines of code xD Why do we need an HTML file that is  243,245 lines long?

roconnor-blockstream:
The `alectryon` directory is a copy of the `nix-build -A coq.doc` output.

Normally it is bad form to check in build outputs, but it quite convenient to point people to

    https://htmlpreview.github.io/?https://github.com/BlockstreamResearch/simplicity/blob/master/alectryon/verif_modinv64_impl.v.html

which allows people to view the context of the Coq proof development without needing to install and run Coq.  Otherwise Coq proof scripts themselves are unreadable because you cannot view the context in which the tactics are being run.

Anyhow, if @apoelstra has opinions on the matter, I'm certainly open to other options.

uncomputable:
I feel like the `alectryon` HTML pages are mostly useful for presentations, but normal people won't look at it. People who are interested will have Coq installed.

Maybe 2.6 MB of extra data is acceptable to download for every person who clones the repo, but we could also move the files to a separate branch, like the PDF branch.

apoelstra:
I think it's fine to commit the giant file.

2.6 MB is a lot relative to text files but in absolute terms it's not much (and I assume it won't change often). I think it's fine.

PR #275, opened 2024-12-05
Fix malforemd Confidential(ONE)

roconnor-blockstream:
Confidential(ONE), and hence its deriviative Asset(ONE) and Amount(ONE), contains a record structure, which isn't part of the Simplicity type system, and hence cannot be used in Simplicity types.

This reworks separate definitions for Asset and AssetWith (and similarly for Amount/AmountWith) such that Asset and AssetWith are proper Simplicity types.


roconnor-blockstream:
@uncomputable A review isn't required, but if you want to review it you can.

uncomputable (approved):
ACK 65d36e6af4009b51f01c46b7302b7e25b9cb0c71 I looked at the PDF and the types make sense now.

Issue #167, opened 2023-07-07
Update "STOP" code in the tech report


roconnor-blockstream:
We have never used it in practice, and now it is being used to encode a "unary" disconnect for reasoning about commitment-time Simplicity programs.

PR #276, opened 2024-12-06
Remove uses of stop codes

roconnor-blockstream:
The encoding with a stop code was never used in practice. Instead that codeword has been reserved for wallet uses as a "unary disconnect" for cases where a wallet want to use the serialization format but doesn't yet have the disconnected expression available.


roconnor-blockstream:
Fixes #167

roconnor-blockstream:
This PR doesn't touch any C code.  Technically it doesn't need review, but you folks can review if you like.

@apoelstra You run your CI at your leisure.

PR #277, opened 2024-12-18
Identity hash

### LINKED ISSUE #150: Clairify definition of Identity Merkle Root
As noted by @uncomputable, the definition of the IMR in the Tech Report is not the same as the definition used in Haskell and C code.  In the Haskell and C code the Tech Report defined IMR is additionally hashed with the input and output types.  This is done to implement the Tech Report requirement that ⟨imr, input type, output type⟩ triples are unique (up to finding hash collisions).  However the resulting terminology is confusing.

roconnor-blockstream:
Renaming of identifiers in C and Haskell code to refer to identity hash instead of identity Merkle root or IMR.

What we had been calling the IMR in this code is actually hash of the triple of the identity merkle root, and the type roots of the input and output types.  The identity merkle root and the types of the input and outputs together identify expressions uniquely, which is why this combination is used in the codebase.  However the naming in the codebase was confusing.

Fixes #150.

### DIFF
--- a/C/ctx8Pruned.c
+++ b/C/ctx8Pruned.c
@@ -270,8 +270,8 @@ const uint32_t ctx8Pruned_cmr[] = {
   0x7f11746fu, 0xb68fdaedu, 0x3cadda80u, 0xc7cd0245u, 0xa341b927u, 0xe98e60f8u, 0x745dc441u, 0xe11ce1a3u
 };
 
-/* The identity Merkle root of the above ctx8Pruned Simplicity expression. */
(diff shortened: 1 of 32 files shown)


apoelstra (approved):
ACK 3fe7d0efceb6f2c11bfc23664ecaf9a3555f2f5b; successfully ran local tests; will need to make corresponding change in rust-simplicity

uncomputable (approved):
ACK 3fe7d0efceb6f2c11bfc23664ecaf9a3555f2f5b

Issue #195, opened 2023-10-31
Autogenerate jet decoding


uncomputable:
It would help with review and reduce potential human error to autogenerate [how the C code decodes jets](https://github.com/BlockstreamResearch/simplicity/blob/7a6ba113dff8d77e65a5f35d2ac613e219633125/C/primitive/elements/primitive.c#L70).

roconnor-blockstream:
Fixed in #278.

PR #278, opened 2025-01-02
Generate contents of decodePrimitive automatically

roconnor-blockstream:
The new generation code isn't exactly the same as the orginal code. There is no optimistic lifting out of a seconde decoding (code2 variable). Such optimizations could be added back in if wanted, but it is probably best to keep the code straightforward.

The decodePrimitive is already being tested by the "Serialization.C" tests in the Haskell test suite via FFI bindings.


roconnor-blockstream:
Fixes #195.

@apoelstra you are going to need to update CI to run `GenDecodeJet` to make sure the content of the repo matches what is checked in.

As noted above, when the generated code is not identical to the original code.  I suggest something like `git diff master --color-moved --color-moved-ws=allow-indentation-change` when reviewing the changes.

uncomputable (approved):
ACK de799bd8ea630dd4399ad3d78223d233fc760810 I manually compared the old and new files and they seem to do the same thing.

apoelstra:
> `git diff master --color-moved --color-moved-ws=allow-indentation-change`

I'll see what I can do. Inside the nix sandbox I don't have the git repo. Though of course I can run `git init` and `git add` and get access to `git-diff` that way.

Having said that, checking this by hand I see:

* `decodeCoreJets.inc` exactly matches `C/decodeCoreJets.inc`
* `decodeElementsJets.inc` does not appear anywhere in the repo. Is this supposed to be committed?

uncomputable:
@apoelstra `decodeElementsJets.inc` lives in `C/primitives/elements`, no?

apoelstra:
Oh, yes, sorry, I was invoking `find` incorrectly.

apoelstra (approved):
ACK de799bd8ea630dd4399ad3d78223d233fc760810; successfully ran local tests

apoelstra:
Updated my local CI with an exact diff check. If this starts failing I will explore ways to make it more flexible.

roconnor-blockstream:
Don't you already have your CI comparing the output of `GenPrecompute` and friends?

apoelstra:
@roconnor-blockstream yes.

Issue #205, opened 2023-11-20
Improve description of SIMPLICITY_ERR_HIDDEN


uncomputable:
https://github.com/BlockstreamResearch/simplicity/blob/5e7976326156ac9db9b5f8695dd597e8e370e130/C/deserialize.c#L46

It is unclear what illegal children are.

Non-case nodes may not have hidden children, so these children are illegal. However, a case node may not have two hidden children, so these children are also illegal. SIMPLICITY_ERR_HIDDEN is returned in both cases.

apoelstra:
HIDDEN children are illegal if they don't occur as part of an assertion.

Agreed, this text should be improved.

roconnor-blockstream:
illegal HIDDEN children is formally defined in https://github.com/BlockstreamResearch/simplicity/blob/5e7976326156ac9db9b5f8695dd597e8e370e130/C/dag.h#L308-L316 (technically that defines legal HIDDEN children).

I suppose I can break the various conditions of well-formedness into their own names.  But I'm not going to copy the entire formal definition into the error message.

Let me see what I can do.

uncomputable:
Maybe "The parent of a hidden node is not an assertion." Assertions are defined to be case nodes with one left or right hidden child.

roconnor-blockstream:
Fixed in #279.

PR #279, opened 2025-01-03
Clarify comment about illegal hidden children

roconnor-blockstream:


roconnor-blockstream:
@uncomputable Does this rewording address your concerns in #205, or does it need further improvement?

apoelstra (approved):
ACK 4ef6fc264e7235485f5ad3b13eaf0746cb7777cc; successfully ran local tests

C/include/simplicity/errorCodes.h
```diff
@@ -65,7 +65,7 @@ static inline const char * SIMPLICITY_ERR_MSG(simplicity_err err) {
   case SIMPLICITY_ERR_STOP_CODE:
     return "Program has STOP node";
   case SIMPLICITY_ERR_HIDDEN:
-    return "Program has illegal HIDDEN child node";
+    return "Program has node with a HIDDEN child in a position where it is not allowed";
```
uncomputable, line 68: 4ef6fc264e7235485f5ad3b13eaf0746cb7777cc: I think the message doesn't tell the developer how to fix the error. Non-case nodes may not have hidden children. Case nodes may have up to one hidden child.

Maybe "Hidden node is not part of a well-formed assertion" This is close to @apoelstra 's proposal from #205

uncomputable: In a Simplicity DAG, a hidden node can have multiple parents. A hidden node can be part of a well-formed assertion and, at the same time, part of a "not well-formed assertion".

I start to like Russell's proposed error message.

uncomputable (approved):
ACK 4ef6fc264e7235485f5ad3b13eaf0746cb7777cc

roconnor-blockstream:
rebased and expanded new error message comments to other instances in the documentation

apoelstra (approved):
ACK e6c505d98d5daeefbeab80b47015f7e3ea8ca8e8; successfully ran local tests

PR #280, opened 2025-01-03
Rename STOP code error message

roconnor-blockstream:
In 3879496ff475b1dcaff4526d9bad4744dc0987d3 references to stop codes were removed since we never used the stop-code based serialization format. Instead this codeword has been reserved for a disconnect node without a connected child.

Such disconnected nodes cannot be used on chain, but can be used for wallet purposes where they may wish to otherwise mostly reuse the onchain serialization format.

In any case, this codeword is now considered a reserved codeword rather than a STOP code.

### DIFF
--- a/C/deserialize.c
+++ b/C/deserialize.c
@@ -42,7 +42,7 @@ static simplicity_err getHash(sha256_midstate* result, bitstream* stream) {
 /* Decode a single node of a Simplicity dag from 'stream' into 'dag'['i'].
  * Returns 'SIMPLICITY_ERR_FAIL_CODE' if the encoding of a fail expression is encountered
  *   (all fail subexpressions ought to have been pruned prior to serialization).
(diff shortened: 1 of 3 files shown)


apoelstra (approved):
ACK 4f29c12409f8229395aca2da5732571dd43c99dc; successfully ran local tests

uncomputable (approved):
ACK 4f29c12409f8229395aca2da5732571dd43c99dc

roconnor-blockstream:
Fixed missing change in header file.

uncomputable (approved):
ACK a322b72f4de15aaedf2ca0e323e238b4deee40ff

apoelstra (approved):
ACK a322b72f4de15aaedf2ca0e323e238b4deee40ff; successfully ran local tests

PR #283, opened 2025-01-15
fix error code parsing in test harness

roconnor-blockstream:
This should have been included as part of 1b85dc31d80d36dc012755e4369aeeac815476a6.


roconnor-blockstream:
I'm not sure of a way of good way of making this more robust against changes.  However `errors.h` shouldn't be changing much in the future, beyond possibly adding new error codes.

apoelstra:
Yeah. We have the same issue in Rust. Not really a good way to deal with it.

uncomputable (commented):
ACK c9e6d54fd61568050ac7bd1fb98293f20323e456

PR #285, opened 2025-01-28
Add explicit deallocation functions to the Simplicity API

### LINKED ISSUE #284: Add deallocation to the Simplicity API
The Simplicity library API has various `simplicity_*_malloc` structures that dynamically allocate memory.  Currently users are expected to free this memory with C's `free` function.  However in #222 the internal memory allocation interface was abstracted so that it could be changed to use a different allocator. e.g. For `rust-simplicity`'s WASM support the C free function isn't available.

However the internal `simplicity_free` function is not part of the API, so, in awkward cases, there can sometimes be a mismatch in applications using the Simplicity library where  `simplicity_*_malloc` are allocated with a custom allocation function but then are incorrectly `free`d using the standard C allocator.

A good way of making everyone happy is to add deallocation functions for each object to the Simplicity API that in turn call's `simplicity_free`, ensuring that the proper deallocator is pair with the same kind of allocated memory.

roconnor-blockstream:
Fixes #284

### DIFF
--- a/C/include/simplicity/elements/env.h
+++ b/C/include/simplicity/elements/env.h
@@ -94,6 +94,10 @@ typedef struct transaction transaction;
  */
 extern transaction* simplicity_elements_mallocTransaction(const rawTransaction* rawTx);
 
+/* Free a pointer to 'transaction'.
+ */
+extern void simplicity_elements_freeTransaction(transaction* tx);
+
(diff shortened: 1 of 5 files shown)


apoelstra (approved):
ACK 955bcc584e4fef1353b1ba5bd97810cf34953f78; successfully ran local tests

PR #287, opened 2025-02-12
Remove unneed UBOUNDED_MAX checks

roconnor-blockstream:
For the case of `COMP`, if `type_dag[COMP_B(...)].bitSize` exceeds `UNBOUNDED_MAX`, then the bounded_add and friends functions will already set `extraCellsBound[0]` and `extraCellsBound[1]` to `UBOUNDED_MAX`.  Therefore the branch doesn't have any computational effect.  It is also not a case worth optimizing for.

The case of `DISCONNECT` is similar, however possibly only one of `extraCellsBound[0]` or `extraCellsBound[1]` will end up being set to `UBOUNDED_MAX` instead of both.

This is fine, because these values are just intermediate values for the `*cellsBound` computation at the end of the function. There we take the bounded_max of both the `extraCellsBound[0]` and `extraCellsBound[1]` functions. So the result is the same whether or not both `extraCellsBound` values are maxed out or only one is.

### DIFF
--- a/C/eval.c
+++ b/C/eval.c
@@ -629,19 +629,12 @@ simplicity_err simplicity_analyseBounds( ubounded *cellsBound, ubounded *UWORDBo
                                                , bound[dag[i].child[1]].cost ));
       break;
      case DISCONNECT:
-      if (UBOUNDED_MAX <= type_dag[DISCONNECT_W256A(dag, type_dag, i)].bitSize ||
-          UBOUNDED_MAX <= type_dag[DISCONNECT_BC(dag, type_dag, i)].bitSize) {
-        /* 'BITSIZE(WORD256 * A)' or 'BITSIZE(B * C)' has exceeded our limits. */
-        bound[i].extraCellsBound[0] = UBOUNDED_MAX;
-        bound[i].extraCellsBound[1] = UBOUNDED_MAX;
-      } else {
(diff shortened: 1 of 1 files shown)


roconnor-blockstream:
### How did we get here

It's worth reviewing how this code came to be before tearing down this "Chesterton's fence".

Originally we didn't compute `extraCellsBound` and only `extraUWORDBound`

https://github.com/BlockstreamResearch/simplicity/blob/ba3e39f9e8e379349dbe5024adf7d502c1dde718/C/eval.c#L711-L722

Back then, for `COMP`, `type_dag[COMP_B(...)].bitSize` was passed to `ROUND_UWORD` and we needed the check to make sure we  had not hit the maximum value yet before doing that.  (And similarly for `DISCONNCECT`).

In b075e2c3e4eac53aa3f60851659a1f255a687ceb we added the computation for `extraCellsBound` to the `extraUWORDBound`, keeping the same `if` structure and keeping the `extraUWORDBound` computation the same.

In 6f3a3e0bb96871284e3a381ed969baff1a5b44e3 we decide that it was fine for `extraUWORDBound` to perform unsigned overflow because it can only overflow if `extraCellsBound` maxes out, and we already check that case for `extraCellsBound`.  See the comment https://github.com/BlockstreamResearch/simplicity/blob/5afc82e6ec895ae22b066f810bb29c6943b7f0ce/C/eval.c#L682-L691

With this change, the whole existence of the `if` branch became redundant.

PR #293, opened 2025-05-16
Elements refactor

roconnor-blockstream:
Add `Elements` into various `struct` tag names so that they will not conflict with similar structures for a forthcoming Bitcoin application.


apoelstra (approved):
ACK 88fdc3b97549c2d423435ff1b2ec9ac9f68e664c; successfully ran local tests

PR #294, opened 2025-05-16
Remove intermedite primitive directory

roconnor-blockstream:
Since refactoring this directory is now otherwise empty.


apoelstra (approved):
ACK 1e65ccd5c9029dd7324e1893c6562f5b4217041a; successfully ran local tests

Issue #290, opened 2025-05-01
Support for exact annex padding


roconnor-blockstream:
Simplicity programs have a cost associated with them.  This cost is computed from a worst case static analysis of the (pruned) Simplicty program prior to execution.  This cost is computed in milliWU and must be less than or equal to the provided budget, which is defined to be the WU of the input under consideration.

We expect that most common and simple use cases of Simplicity, the budget computed form the serialized Simplicity program will be more than adequate to cover the cost computed form the static analysis of the program.

However, the due to sub-expression sharing, the cost of Simplicity programs can be exponential in the serialized size of the program, leading to cases where the budget is inadequate to cover the required costs.  In particular this can happen when using the higher level `for_while` construction.  The intended solution in this case is to pad the input's annex in order to increase the budget to adequate levels.

However, there is a design flaw with Taproot's annex, an input's signature can only cover it's own annex, and not the annex of other inputs.
While most of the time users are incentivized to minimize their transaction's weight, in some circumstances within a multi-party protocol, one party could prefer to sabotage the transaction they are participating in by substituting their input's annex with a very large annex in order to increase the transaction size and decrease the transaction's fee weight.  Because inputs only sign their own annex, this substitution can even be done after the transaction is signed by all parties, and one party replaces their own signature with an otherwise identical one signing a heavy annex.  This party would then try to quickly broadcast this heavy transaction, and "pinning" the other party's inputs.

With fullRBF on by default, such pinning attacks are perhaps harder to pull off, but I believe it is still a concern.

Currently we have standardness rules that forbid the annex form appearing in transaction, and these standardness rules prevent such attacks from being practical to pull off.  However, this standardness rules rule currently forbids padding the annex in cases where it is needed to increase the budget of Simplicity programs.

To this end we want to carve out a minimal standardness rule to allow padding of Simplicity programs, at least until we get other adequate rules regarding allowed annex uses.  The obvious minimal standardness rule is an annex consisting of only  `0x00` bytes that is adequate to cover the costs of an input's Simplicity program, but no more.  The carve out only allows a single possible annex for simplicity programs, one that is adequate for satisfying it's needed budget.

[...]

PR #295, opened 2025-05-16
Add minCost parameter

roconnor-blockstream:
This is designed to support exact annex padding standardness rule by returning an OVERWEIGHT condition when the computed cost is below some minCost.

Setting minCost to 0 effectively nullifies this check.

Technically setting the minCost rejects programs that have a computed cost of 0, however there are no Simplicity programs that have a computed cost of 0 since every Simplicity program must contain at least one node, and hence has at least 'overhead' cost, which is 100 milliWU.

### DIFF
--- a/C/eval.c
+++ b/C/eval.c
@@ -571,6 +571,7 @@ typedef struct boundsAnalysis {
  * When maxCells < UBOUNDED_MAX, if the bounds on the number of cells needed for evaluation of 'dag' on an idealized Bit Machine exceeds maxCells,
  * then return SIMPLICITY_ERR_EXEC_MEMORY.
  * When maxCost < UBOUNDED_MAX, if the bounds on the dag's CPU cost exceeds 'maxCost', then return SIMPLICITY_ERR_EXEC_BUDGET.
+ * If the bounds on the dag's CPU cost is less than or equal to 'minCost', then return SIMPLICITY_ERR_OVERWEIGHT.
  * Otherwise returns SIMPLICITY_NO_ERR.
  *
  * Precondition: NULL != cellsBound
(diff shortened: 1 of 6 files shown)


C/eval.c
```diff
@@ -741,15 +740,17 @@ simplicity_err simplicity_analyseBounds( ubounded *cellsBound, ubounded *UWORDBo
    */
   return (maxCells < *cellsBound) ? SIMPLICITY_ERR_EXEC_MEMORY
        : (maxCost < *costBound) ? SIMPLICITY_ERR_EXEC_BUDGET
+       : (*costBound <= minCost) ? SIMPLICITY_ERR_OVERWEIGHT
```
apoelstra, line 743: In d2c2de51882ad95ab5618f87582127e3e4f18790:

I think we want this `<=` to be a `<`. Then both `minCost` and `maxCost` are inclusive bounds, which I think is what's implied by their names.

apoelstra:
In d2c2de51882ad95ab5618f87582127e3e4f18790:

I'd also like to add a simple test, one which just takes one of the test programs and checks that it's accepted if you set `minCost`/`maxCost` to the actual computed cost, and rejected if they're both moved 1 up or 1 down.

roconnor-blockstream:
fixes #290

C/eval.c
roconnor-blockstream: I'm okay with renaming these fields, but as per the design in #290, the ranges must be non-overlapping:

> Notice that the strictness of the inequalities is important if we are being exact. If the rule were instead l <= c <= b then if it were to happen that c were exactly l down to the the exact milliWU, there would be two annex sizes that would support that cost, which would add a small malleability in the allowed size of the annex.

apoelstra: In the code you allow `l == b` while in #290 you sorta imply, but don't explicitly say, that `l` would be `b - 1` except that because of encoding edge cases sometimes it'll have to be `b - `2.

You could have just as easily used non-strict inequalities, and said that `l` would be `b`, except that because of encoding edge cases sometimes it'll have to be `b - 1`. So we are redefining `l`.

I'm fine with keeping the design as-is, matching #290, but if we do then I definitely think we should rename `minCost`. It could be called `minCostMinusOne`, or `maxInvalidCost`, or something. Hopefully you can think of a less ugly name.

roconnor-blockstream: I allow l == b in the sense that it isn't precondition violating to pass those arguments, however if you do pass arguments such that l == b then there is no code-path that will return `SIMPLICTY_OK`; it will always at least either return `SIMPLICITY_ERR_EXEC_MEMORY` or `SIMPLICITY_ERR_OVERWEIGHT` (or possibly some other error if you have like type errors and such that are found first).

I allow `l == b` so that degenerate requests can be made without violating my precondition.

That said, I'm happy to drop that precondition and just double check that everything remains sensible when the user passes b < l.

> You could have just as easily used non-strict inequalities, and said that l would be b, except that because of encoding edge cases sometimes it'll have to be b - 1.

Unfortunately no.  The issues is that the API takes mincosts and maxcosts in terms of WU, but the weight calculations are done in milliWU.  Obviously subtracting 1 WU is not the same as subtracting 1 milliWU.

The conversion from WU to milliWU is done on line 386.  Maybe I should change variable names to note their units to make this more apparent.

apoelstra: Ok, I understand!

Can you add a comment explaining that both values are in milliWU, but the granularity of the cost bounds will be in wu, where wu *n* is represented by the half-open interval (1000 * (n - 1), 1000 * n], and that we therefore check `minCost` using a strict inequality and `maxCost` using a `<=` inequality.

I think that the existing names are fine as long as you provide that intuition pump.

roconnor-blockstream:
> I'd also like to add a simple test, one which just takes one of the test programs and checks that it's accepted if you set minCost/maxCost to the actual computed cost, and rejected if they're both moved 1 up or 1 down.

I dismissed this earlier because finding an example program that is exactly a multiple of 1000 milliWU.  However, since you bring it up, perhaps it is better for us to find and test such a program now, than for our adversaries to find it later.

roconnor-blockstream:
Updated.  Let me know what you think of the add comments.

apoelstra:
bb3f0396dda54270dbb86583f81c09f55452b421 looks good! Except that in the last commit it looks like you added a file called `-i`.

apoelstra:
Heh, yeah. Definitely takes some extra steps to get your branches into a publishable state.

apoelstra (approved):
ACK 124b3190deba29fa500ab6cc597ae2a985588820; successfully ran local tests

PR #297, opened 2025-05-22
Fixup cabal repl Simplicity

roconnor-blockstream:
Recent refactoring in 1a4006cc260754919ba865ca3c8d2f5f4de32df8 inserted txEnv.c into the wrong place. This caused `cabal repl Simplicity` to fail.

### DIFF
--- a/Simplicity.cabal
+++ b/Simplicity.cabal
@@ -124,7 +124,7 @@ library Simplicity-Indef
                        vector >=0.12 && <0.14
 
 library
-  C-sources:           C/rsort.c C/elements/jets.c C/elements/ops.c C/elements/env.c
+  C-sources:           C/rsort.c C/elements/jets.c C/elements/ops.c C/elements/env.c C/elements/txEnv.c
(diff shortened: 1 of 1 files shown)


roconnor-blockstream:
@apoelstra I don't really know a good way of running `cabal repl Simplicity` in your CI.  So I'm inclined just to fix this on an ad hoc basis.

delta1:
Might be time for some github CI? I can take a look, what kind of test jobs should be run?

roconnor-blockstream:
@apoelstra has in own CI in NixOS, whose details we should probably merge into this project itself.

apoelstra:
My local CI is pretty heavy. I'm not sure it's reasonable to merge into this project.

But if you want it it's here: https://github.com/apoelstra/local-nix-ci/blob/main/BlockstreamResearch.Simplicity.check-pr.nix

delta1:
Thank I’ll take a look!

roconnor-blockstream:
rebased

PR #298, opened 2025-06-19
Rename elements/jets.c

roconnor-blockstream:
I forgot that MSVC doesn't let you have files with the same name, even if they are in different directories.


delta1 (approved):
utACK dd11891 

windows ci task timeout is unrelated, so this does fix the msvc build issue

apoelstra (approved):
ACK cad9a63292ea8fdd5c1615e0322986f6f996d42f; successfully ran local tests

PR #309, opened 2025-08-21
Add Dockerfiles for building and running C,Coq and Haskell projects

imaginator:
I've created the following files:

```
Dockerfile.c
Dockerfile.haskell
Dockerfile.coq
```

Builds are now self contained and drops the user into either a `ghci` or `coq` prompt. Ideally we could publish these containers to Dockerhub which would make it easier for experimenting [without waiting for long builds]. 

The README.md needs a bit of a rework too, but that's another project.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+.dockerignore
+.git
+docker-compose.yml
+Dockerfile
+Dockerfile.c
+Dockerfile.coq
+Dockerfile.haskell
--- a/.gitignore
+++ b/.gitignore
@@ -33,5 +33,6 @@ alectryon-doc
 # Gen
 /*.rs
 /*.c
+!/Dockerfile.c
 /*.h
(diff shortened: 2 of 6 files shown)


imaginator:
@apoelstra their use is documented in the `README.md` file. For example the building and running the Coq version should give a `coqtop` prompt with everything built. Likewise, the haskell image presents a GHCi console with correct Simplicity modules loaded. (https://github.com/BlockstreamResearch/simplicity/pull/309/commits/6892b08658a9ddfca7a47f11372ce0e00c7d9a1b)

My thinking behind these is that it makes an easy way to write tutorials and, like with Nix images, present users with all the dependencies they need to get experimenting with Simplicity quickly.

apoelstra (approved):
ACK 6892b08658a9ddfca7a47f11372ce0e00c7d9a1b; successfully ran local tests

PR #321, opened 2025-12-05
Bitcoin jets

roconnor-blockstream:
Implement our set of jets for the Bitcoin application of Simplicity.

This also includes defining the transaction environment for the Bitcoin application of Simplicity.


roconnor-blockstream:
Coinbase transactions are not allowed to have extra inputs.

Regardless the specification states the fee is taken modulo 2^64, which is well-defined.  There is plenty of room in 64 bits to consider values in the range [-money_range, +money_range] if users need to interpret the result as a signed 64 bit value.

PR #326, opened 2025-12-26
Depricate broken relative timelock jets

### LINKED ISSUE #325: Relative timelock jets have broken semantics
I very incorrectly computed the max of the all relative timelock values for comparison when specifying the check relative timelock jets.

We need to fix this by 

(1) renaming these broken jets to dissuade their use in Elements.

(2) add fixed versions for the Bitcoin application.

roconnor-blockstream:
Fixes #325.


delta1 (approved):
ACK a97e57cd96f1ae110d75429043393fca10702a3e 

Ran nix-build for C and Haskell locally

PR #334, opened 2026-03-19
Update Rust code generation to produce the jet environment separately

ivanlele:
Generates a jet implementation for the updated jet trait introduced in BlockstreamResearch/rust-simplicity#351


apoelstra (approved):
ACK 2ee52e3cee4ce1b5c394c09f29f05b589b40af17; successfully ran local tests

PR #339, opened 2026-06-03
feat: add output op_return get_bytes jets

stringhandler:
Adds `output_null_get_bytes_{1,2,4,8,16,32,64}` jets for reading raw byte data from null-data (OP_RETURN) outputs, and the supporting `OP_PUSHBYTES_*` opcode variants that make them possible.

## Background

Previously, `output_null_datum` could identify a push in a null-data script and return its SHA-256 hash, but there was no way for a Simplicity program to access the actual bytes. This matters for null-data outputs that carry small, fixed-size payloads (nonces, amounts, asset IDs, public keys, signatures) that a contract needs to inspect directly.

**New jets (`C/elements/elementsJets.c`)**

`simplicity_output_null_get_bytes_{1,2,4,8,16,32,64}` - each takes an output index and a push index and returns `S (S TWO^N)`: nothing if the output isn't null-data, nothing if the indexed push isn't exactly N bytes, or the raw bytes.


apoelstra:
We first need to develop some sort of versioning mechanism so that the Simplicity interpreter knows whether these jets are active or not. Directly enabling them like this would hardfork the Liquid network.

stringhandler:
Yeah, there is the version number in the tapleaf, but we may want to be sparing with version upgrades.

Do you think we need a version for https://github.com/BlockstreamResearch/rust-simplicity/pull/345 ?

stringhandler:
I did add this issue to discuss it https://github.com/BlockstreamResearch/simplicity/issues/341, mainly to catalogue my thoughts, but we can brainstorm it there maybe.

apoelstra:
> we may want to be sparing with version upgrades.

In the 8 years Liquid has been deployed, we have literally never used any of the multiple script upgrade mechanisms that exist, except when we deployed Taproot. If we find ourselves running low on version numbers there are lots of alternate schemes we could deploy, but I'm skeptical that we will.

>Do you think we need a version for BlockstreamResearch/rust-simplicity#345 ?

No. That is just fixing rust-simplicity to support functionality that was already deployed.

stringhandler:
I've added https://github.com/stringhandler/simplicity/pull/1 as a possibly versioning solution. I'll give it some more thought though.

PR #346, opened 2026-07-20
Haskell-Generate: stop referring to `Midstate` explicitly

apoelstra:
This is an unnecessary encapsulation violation (Haskell-Generate doesn't need to know how Cmr is implemented when it just needs to construct one from a byte array), and also it'll stop working when we move to bitcoin_hashes 1.0.

### DIFF
--- a/Haskell-Generate/GenRustJets.hs
+++ b/Haskell-Generate/GenRustJets.hs
@@ -157,7 +157,7 @@ rustJetCmr mod = vsep $
         ]))
     , "};"
     , mempty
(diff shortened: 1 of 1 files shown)


delta1 (approved):
ACK 90b0dbe9b7efd3e1a64c016be52561102f7f5bf3; tested locally

apoelstra:
On 90b0dbe9b7efd3e1a64c016be52561102f7f5bf3 successfully ran local tests

